PR #1, opened 2016-03-31
Create tcp_timestamps.conf

HulaHoopWhonix:


adrelanos:
TODO:

* It needs also a `Replaces: tcp-timestamps-disable` in `debian/control` to avoid breaking the package management during upgrade. (A file may not be owned by two packages at the same time.) 
* Migrate tcp-timestamps-disable debian/control description.
* Remove tcp-timestamps-disable from anon-meta-packages.
* Remove tcp-timestamps-disable from Whonix git modules.
* Remove tcp-timestamps-disable from Whonix packages folder.

PR #2, opened 2017-03-14
Update README.md

HulaHoopWhonix:


adrelanos:
More important to update debian/control since readme is auto generated from debian/control.

PR #5, opened 2019-05-05
More kernel hardening

madaidan:
Adds more kernel hardening. See https://forums.whonix.org/t/kernel-hardening/7296


adrelanos:
//cc @marmarek

PR #7, opened 2019-05-16
Disable uncommon network protocols

madaidan:
Disables DCCP, SCTP, RDS and TIPC in case they have unknown vulnerabilities. They have had bad vulnerabilities in the past.


adrelanos:
https://github.com/Whonix/security-misc/pull/7

PR #10, opened 2019-06-23
Enable more kernel hardening parameters

madaidan:


adrelanos:
//cc @marmarek

See also:

https://github.com/Whonix/security-misc/commit/807ac7d65916071e4294f42d62b8b2353255c4bc

PR #13, opened 2019-06-25
Remove System.map and restrict the SysRq key.

madaidan:


usr/lib/security-misc/remove-system.map
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# Removes the System.map files as they are only used for debugging or malware.
+if [ -f /boot/System.map-* ]; then
```
adrelanos: When there are multiple files matching this an error

    ./a: line 4: [: too many arguments

will happen.

madaidan: `-f` doesn't work with globs so the error occurs. 8ef0db1 should fix it.

usr/lib/security-misc/remove-system.map
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
 
 # Removes the System.map files as they are only used for debugging or malware.
-if [ -f /boot/System.map-* ]; then
-  rm /boot/System.map-*
-fi
+for file in /boot/System.map-*
```
adrelanos: file is a linux command line tool too, could you use `filename` or so please?

usr/lib/security-misc/remove-system.map
```diff
@@ -1,6 +1,9 @@
 #!/bin/bash
```
adrelanos: Could you please add

    shopt -s nullglob

Then it won't show an error message when there is no file.

Also could you add licensing please?

usr/lib/security-misc/remove-system.map
```diff
@@ -1,6 +1,9 @@
...
 
 # Removes the System.map files as they are only used for debugging or malware.
-if [ -f /boot/System.map-* ]; then
-  rm /boot/System.map-*
-fi
+for file in /boot/System.map-*
+do
+  if [ -f "${file}" ]; then
+    rm "${file}"
```
adrelanos: `rm -f` better? To avoid race condition errors?

usr/lib/security-misc/remove-system.map
madaidan: Is  8ef0db1 alright? I've never used `shopt` before so I'm not sure.

adrelanos: Yes.

Could you try to re-run the script a few times please? Is it idempotent, i.e. there will be no errors if it gets re-run?

madaidan: It is. There are no errors if it gets re-run. I've tried it multiple times.

PR #16, opened 2019-06-29
Mount /proc with hidepid=2

madaidan:


lib/systemd/system/proc-hidepid.service
```diff
@@ -0,0 +1,27 @@
+[Unit]
```
adrelanos: Could you please add licensing comments on the very top?

madaidan: Done.

https://github.com/Whonix/security-misc/pull/19

PR #26, opened 2019-07-21
Fix files

adrelanos:


adrelanos:
Cool, thanks!

Could you please test/port to %files using the wildcard `/*` so this list does not have to be manually maintained?

adrelanos:
(That was discussed and OK'd in https://github.com/QubesOS/qubes-issues/issues/1885 too.)

fepitre:
Yes sure. I will have a look on that. I saw the discussion indeed and if I can handle a simple extract from debian control file for example, I would do it in preference from ```/*```.

PR #31, opened 2019-10-03
Restrict /proc/cpuinfo, /proc/bus, /proc/scsi and /sys to root

madaidan:
This hides hardware information to unprivileged users and increases security. See https://forums.whonix.org/t/anonymize-etc-machine-id/7721/21 and https://forums.whonix.org/t/restrict-hardware-information-to-root/7329/

### DIFF
--- a/lib/systemd/system/hide-hardware-info.service
+++ b/lib/systemd/system/hide-hardware-info.service
@@ -1,3 +1,6 @@
+## Copyright (C) 2012 - 2018 ENCRYPTED SUPPORT LP <adrelanos@riseup.net>
+## See the file COPYING for copying conditions.
+
 [Unit]
 Description=Hide hardware information to unprivileged users
 Documentation=https://github.com/Whonix/security-misc
--- a/lib/systemd/system/hide-hardware-info.service
+++ b/lib/systemd/system/hide-hardware-info.service
@@ -1,0 +1,14 @@
+[Unit]
+Description=Hide hardware information to unprivileged users
+Documentation=https://github.com/Whonix/security-misc
+DefaultDependencies=no
+Before=sysinit.target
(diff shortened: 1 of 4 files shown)


lib/systemd/system/hide-hardware-info.service
```diff
@@ -0,0 +1,17 @@
...
+Before=sysinit.target
+Requires=local-fs.target
+After=local-fs.target
+
+[Service]
+Type=oneshot
+ExecStart=/usr/lib/security-misc/hide-hardware-info
+
+[Install]
+WantedBy=multi-user.target
```
adrelanos: `WantedBy=multi-user.target` is probably wrong.

Could you try this please:

    WantedBy=sysinit.target

`WantedBy=multi-user.target` is probably wrong in other systemd unit files too and this mistake is now propagating.

adrelanos: To find other examples for systemd early boot services:

    grep -r WantedBy /lib/systemd/system | grep sysinit

madaidan: > `WantedBy=multi-user.target` is probably wrong.

It worked fine when I tested it.

More services use `multi-user.target` than `sysinit.target`.

Compare

    grep -r "WantedBy=multi-user.target" /lib/systemd/system | wc -l

to

    grep -r "WantedBy=sysinit.target" /lib/systemd/system | wc -l

adrelanos:
It's not a vote. It depends on what the systemd unit file does. All that
use DefaultDependencies=no probably use sysinit.target.

madaidan:
Alright. Now changed to sysinit.target.

https://github.com/madaidan/security-misc/commit/7345287560bc701f8b4aead985238d66104b228c

PR #32, opened 2019-10-04
Disable TCP DSACK and FACK

madaidan:
As per https://forums.whonix.org/t/disabling-tcp-sack-dsack-fack/8109


adrelanos:
Will merge and delete later as per:
 https://forums.whonix.org/t/disabling-tcp-sack-dsack-fack/8109/5

PR #41, opened 2019-12-07
Check for more locations of System.map

madaidan:
https://forums.whonix.org/t/kernel-hardening/7296/264

### DIFF
--- a/usr/lib/security-misc/remove-system.map
+++ b/usr/lib/security-misc/remove-system.map
@@ -5,8 +5,10 @@
 
 shopt -s nullglob
 
+system_map_location="/boot/System.map* /usr/src/*/System.map* /lib/modules/*/*/System.map* /System.map*"
+
 counter=0
-for filename in /boot/System.map* ; do
+for filename in ${system_map_location} ; do
    counter=$(( counter + 1 ))
 done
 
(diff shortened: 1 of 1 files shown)


usr/lib/security-misc/remove-system.map
```diff
@@ -5,8 +5,10 @@
 
 shopt -s nullglob
 
+system_map_location="/boot/System.map* /usr/src/*/System.map* /lib/modules/*/*/System.map* /System.map*"
```
adrelanos: Did you test this? I guess this possibly won't work. The `*` inside quotes `"`  is just a character. Does not expand.

Or does it expand later in `${system_map_location}` since that does not use quotes?

madaidan: > Or does it expand later in ${system_map_location} since that does not use quotes?

Yes it does.

Running:

```
for i in ${system_map_location}; do   echo $i; done
```

on my host outputs:

```
/boot/System.map*
/usr/src/linux-hardened/System.map
/usr/src/linux/System.map
/lib/modules/5.3.13.a-1-hardened/build/System.map
/lib/modules/5.3.13-arch1-1/build/System.map
/System.map*
```

so it does expand.

The `*` only becomes a character when the files don't exist which isn't a problem as we check if the files do exist further down.

PR #43, opened 2019-12-18
More kernel hardening

madaidan:


adrelanos:
There were some merge conflicts so I manually copied the file. Please check.

PR #46, opened 2019-12-21
Don't remount /sys/kernel/security

madaidan:
It's mounted with those options by default anyway

### DIFF
--- a/usr/lib/security-misc/remount-secure
+++ b/usr/lib/security-misc/remount-secure
@@ -101,11 +101,6 @@ _tmp() {
    remount_secure "$@"
 }
 
-_sys_kernel_security() {
-   new_mount_options="nosuid,nodev${noexec_maybe}"
-   remount_secure "$@"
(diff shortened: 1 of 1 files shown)


adrelanos:
I've taken care of this otherwise: if correct mount options are already set, it will not remount.

https://github.com/Whonix/security-misc/blob/master/usr/lib/security-misc/remount-secure#L60-L62

Therefore this helps on systems without apparmor-profile-everything.

And when using apparmor-profile-everything this pull request isn't needed?

madaidan:
No, it's the default regardless if apparmor-profile-everything is used.

The only reason I suggested to mount securityfs with these mount options in the first place is because https://github.com/Whonix/apparmor-profile-everything/blob/master/etc/initramfs-tools/scripts/init-bottom/apparmor-profile-everything mounted it with no options at all resulting in worsened security from without using apparmor-profile-everything.

PR #51, opened 2019-12-24
Don't display kernel logs during boot

madaidan:
https://phabricator.whonix.org/T950


adrelanos:
Potentially better implementation as per https://phabricator.whonix.org/T950#19230 needs to be researched.

PR #53, opened 2020-01-12
Set sysctl values in initramfs

madaidan:
https://forums.whonix.org/t/kernel-hardening/7296/331


adrelanos:
https://github.com/Whonix/security-misc/pull/55

PR #69, opened 2020-03-07
Use systemd .mount files for stricter mount options and remove remount-secure

madaidan:
The only missing parts of this are the /tmp, /dev/shm and /run restrictions as systemd doesn't seem to like them:

    dev-shm.mount: Cannot create mount unit for API file system /dev/shm


adrelanos:
https://forums.whonix.org/t/re-mount-home-and-other-with-noexec-and-nosuid-among-other-useful-mount-options-for-better-security/7707/42

adrelanos:
https://forums.whonix.org/t/re-mount-home-and-other-with-noexec-and-nosuid-among-other-useful-mount-options-for-better-security/7707/45

adrelanos:
Stalled.

PR #74, opened 2020-03-30
debian: .preinst, add user to groups automatically

h0tw4t3r:
Propose user to add himself to sudo and console group automatically.


h0tw4t3r:
Yeah, you definitely right, let me fix this up.

h0tw4t3r:
Hey, can I have Whonix apt source link?

adrelanos:
The fully complete source code for security-misc can be found in this repository where you're already commenting:
https://github.com/Whonix/security-misc

No source code for security-misc needs to be downloaded from deb.whonix.org.

----

> I need to `apt source security-misc`

Not useful but anyhow...

File `/etc/apt/sources.list.d/whonix.list`.

> deb tor+https://deb.whonix.org buster-developers main contrib non-free
> #deb-src tor+https://deb.whonix.org buster-developers main contrib non-free

Comment in the deb-src line. I.e. complete example:

> deb tor+https://deb.whonix.org buster-developers main contrib non-free
> deb-src tor+https://deb.whonix.org buster-developers main contrib non-free

Also feel free to remove the `tor+` part if you're not using apt-transport-tor / Tor. I.e. complete example:

> deb https://deb.whonix.org buster-developers main contrib non-free
> deb-src https://deb.whonix.org buster-developers main contrib non-free

Then you can download the source package.

    sudo apt update
    apt source security-misc

But I don't see much point in that. Better get sources from git.

h0tw4t3r:
Sorry, I am not a professional Debian package maintainer, but thanks for helping me out.

I wanted `apt source` so I could have the `orig` and `debian` tarballs.
Also github repo's version is `17.0`, and the one on the repos is `16.8`.
As I am doing changes to preinst and config files I really wanted to check them in works, and the easiest way imo was sourcing the package.

However, I tried using it, but `db_get` returns nothing from database. I followed http://manpages.ubuntu.com/manpages/trusty/en/man7/debconf-devel.7.html.

If you ever worked with it would you please be so kind to help me find the bug?

adrelanos:
Rebase required. Probably easiest to just re-add your changes to the preinst.

----

I've just now tested adding `env` on top of the preinst. It contains:

* `SUDO_GID=1000`
* `SUDO_USER=user`
* `USER=root`

> Also github repo's version is `17.0`, and the one on the repos is `16.8`.

Repo version much more important.

> As I am doing changes to preinst and config files I really wanted to check them in works, and the easiest way imo was sourcing the package.

I don't think apt source helps.

Yesterday I simplified building the package. Standard Debian tools can now be easily used such as:

    dpkg-buildpackage -b

Then install (version number is likely to change):

    sudo dpkg -i ../security-misc_17.5-1_all.deb

> However, I tried using it, but `db_get` returns nothing from database. I followed http://manpages.ubuntu.com/manpages/trusty/en/man7/debconf-devel.7.html.
> 
> If you ever worked with it would you please be so kind to help me find the bug?

No. I never used that. Try Debian mentors mailing list.

https://lists.debian.org/debian-mentors/

> And question, are there any throwbacks if we just add current user to those groups automatically, since he already uses sudo to install the package?

That might resolve part of what you want to accomplish here. But there might be a better solution.

Why don't we allow group members of group `sudo` to use console by editing [`/etc/security/access-security-misc.conf`](https://github.com/Whonix/security-misc/blob/master/etc/security/access-security-misc.conf)? Group members of group `sudo` could run `sudo addgroup user console` anyhow. Therefore no security advantage by not doing that.

adrelanos:
Did that. See above comment. Please test.

Is there anything else missing here?

> Propose user to add himself to sudo and console group automatically.

Why would we add someone capable to install packages - a sudo user - to group sudo?

h0tw4t3r:
> Why would we add someone permitted to install packages - a sudo user - to group sudo?

Well atm if there are no users in sudo group, you force current user to add them by hands.
Since the user uses `sudo` to use `apt` already, doesn't this indicate that the user has to be in the sudo group?

Let me get rid of `debconf` libraries and rebase repo.

adrelanos:
> > Why would we add someone permitted to install packages - a sudo user - to group sudo?
> 
> Well atm if there are no users in sudo group, you force current user to add them by hands.

If there's no sudo user when security-mic, then which user is it who has permission to run APT?

> Since the user uses `sudo` to use `apt` already, doesn't this indicate that the user has to be in the sudo group?

If for example environment variable `SUDO_USER=user` is set that indicates that user `user` is already a member of group `sudo`. Then it's pointless and does nothing to add that user `user` to group `user` since that is already the case anyhow.

Which non-root user might be running "apt install"? Then which user could be added?

If there is no sudo user, then in most cases it would be root who is running apt-get, right?

Well, perhaps a user was using `su` to get root privileges and install security-misc. In that case we could use environment variable `USER` which is set for me to `USER=user` after using `su`.

I don't thin we should use environment variable `LOGNAME` since that is empty in my Qubes VM. `LOGNAME` could be tested as a fallback.

Finding out the user name which runs apt should be outsourced to its own shell function. Also check that `root` won't be returned as an answer or if `root` is returned as an answer, then we need to act accordingly. I.e. error out since then we have no idea which user to auto add to group `sudo`.

In Whonix build script I invented this:

```
user_name_detection() {
   ## Since the script runs as root, we can not use $USER or whoami (that would be set to root).
   if [ ! "$user_name" = "" ]; then
      return 0
   fi
   if [ ! "$SUDO_USER" = "" ]; then
      true "INFO $BASH_SOURCE: Setting user_name to SUDO_USER $SUDO_USER..."
      user_name="$SUDO_USER"
   fi
   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
      true "INFO $BASH_SOURCE: Trying to set user_name by using \`logname\`."
      user_name="$(logname)" || true
   fi
   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
      echo "INFO $BASH_SOURCE: SUDO_USER: $SUDO_USER"
      true "INFO $BASH_SOURCE: Trying to set user_name by using \`whoami\`."
      user_name="$(whoami)"
      echo "INFO $BASH_SOURCE: Set user_name to $user_name using whoami."
   fi
   if [ "$user_name" = "" ]; then
      error "Variable user_name is empty."
   fi
}
```

Could use some review / adjustment / improvement.

debian/security-misc.preinst
```diff
@@ -87,21 +87,15 @@ sudo_users_check () {
    IFS="$OLD_IFS"
    export IFS
 
+   ## Prevent users from locking themselves out.
+   ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/4
    if [ "$are_there_any_sudo_users" = "yes" ]; then
+      echo "$0: INFO: Consider next users are in sudo group: $sudo_users"
```
adrelanos, line 93: Kinda an information leak?

debian/security-misc.preinst
```diff
@@ -176,7 +167,10 @@ console_users_check() {
       return 0
    fi
 
-   exit 201
+   echo "$0: WARNING: No user is a member of group 'console'. Adding current user $(logname) to console group." >&2
+   addgroup $(logname) console
```
adrelanos, line 171: This fails for me on Qubes.

    logname

> logname: no login name

    echo $?

>1

And due to the non-zero exit code of `logname` the preinst script would exit non-zero and the apt would report package installation error.

debian/security-misc.preinst
```diff
@@ -87,21 +87,15 @@ sudo_users_check () {
    IFS="$OLD_IFS"
    export IFS
 
+   ## Prevent users from locking themselves out.
+   ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/4
    if [ "$are_there_any_sudo_users" = "yes" ]; then
+      echo "$0: INFO: Consider next users are in sudo group: $sudo_users"
+      echo "$0: INFO: If you want to manage permissions differently, consider running:"
+      echo "sudo adduser user sudo" >&2
```
adrelanos, line 95: Not `>&2`?

debian/security-misc.preinst
adrelanos: There might be cases where we cannot find out any user name because the user logged in as `root`. And root login is disabled by security-misc. Therefore there will still be cases where we need to error out.

debian/security-misc.preinst
h0tw4t3r: Yep, my bad :P

debian/security-misc.preinst
h0tw4t3r: Ok, I will remove these messages.

debian/security-misc.preinst
h0tw4t3r: I will include your `user_name_detection` function that you attached below.
And as I understood there are no confident ways to always have a user's name, so I will return the error throwing part.

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
 #####################################################################
 "
 
+user_name_detection() {
```
adrelanos, line 18: This code wasn't ready for unmodified copy/paste.

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
...
+   ## Since the script runs as root, we can not use $USER or whoami (that would be set to root).
+   if [ ! "$user_name" = "" ]; then
+      return 0
+   fi
+   if [ ! "$SUDO_USER" = "" ]; then
+      true "INFO: Setting user_name to SUDO_USER $SUDO_USER..."
+      user_name="$SUDO_USER"
```
adrelanos, line 25: Must not be done if SUDO_USER is `user`.

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
...
+   fi
+   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
+      true "INFO: Trying to set user_name by using \`logname\`."
+      user_name="$(logname)" || true
```
adrelanos, line 29: Again, don't accept value of `user`.

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
...
+   fi
+   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
+      echo "INFO: SUDO_USER: $SUDO_USER"
+      true "INFO: Trying to set user_name by using \`whoami\`."
+      user_name="$(whoami)"
```
adrelanos, line 34: Again, don't accept value of `user`.

adrelanos: Needs `|| true`.

debian/security-misc.preinst
```diff
@@ -87,12 +111,20 @@ sudo_users_check () {
...
    if [ "$are_there_any_sudo_users" = "yes" ]; then
       return 0
    fi
 
-   ## Prevent users from locking themselves out.
-   ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/4
+   user_name_detection
+   if [ "$user_name" != "" ]; then
+      echo "$0: WARNING: No user is a member of group 'sudo'. Adding current user $user_name to sudo group." >&2
+      addgroup $user_name sudo
```
adrelanos, line 123: Must be quoted. `"$user_name"`

debian/security-misc.preinst
```diff
@@ -159,6 +191,14 @@ console_users_check() {
       return 0
    fi
 
+   user_name_detection
+   if [ "$user_name" != ""]; then
+      echo "$0: WARNING: No user is a member of group 'console'. Adding current user $user_name to console group." >&2
+      addgroup $user_name console
```
adrelanos, line 197: Must be quoted. `"$user_name"`

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
...
+      user_name="$(logname)" || true
+   fi
+   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
+      echo "INFO: SUDO_USER: $SUDO_USER"
+      true "INFO: Trying to set user_name by using \`whoami\`."
+      user_name="$(whoami)"
+      echo "INFO: Set user_name to $user_name using whoami."
+   fi
+   if [ "$user_name" = "" ]; then
+      error "Variable user_name is empty."
```
adrelanos, line 38: Not good to do that in this script.

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
 #####################################################################
 "
 
+user_name_detection() {
+   ## Since the script runs as root, we can not use $USER or whoami (that would be set to root).
```
adrelanos, line 19: Non-applicable comment. Below we actually do use whoami.

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
...
+   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
+      true "INFO: Trying to set user_name by using \`logname\`."
+      user_name="$(logname)" || true
+   fi
+   if [ "$user_name" = "" ] || [ "$user_name" = "root" ]; then
+      echo "INFO: SUDO_USER: $SUDO_USER"
+      true "INFO: Trying to set user_name by using \`whoami\`."
+      user_name="$(whoami)"
+      echo "INFO: Set user_name to $user_name using whoami."
+   fi
```
adrelanos, line 36: Please also check `$USER` (in case of someone using `su` without `sudo`).

debian/security-misc.preinst
adrelanos: Yes. In some cases there is nobody to add. For example on servers it's common to login as root and to run apt under root. In that case we can do nothing except the existing error out in the preinst script.

debian/security-misc.preinst
h0tw4t3r: Ok, as we already check for `user_name`, I will just remove it.

debian/security-misc.preinst
h0tw4t3r: What do you mean by that? We already check if user is empty two lines above, `[ "$SUDO_USER" = "$user_name" ]` is redundant.

adrelanos: If environment Valeria `SUDO_USER` is set to `root`, don't use `SUDO_USER`. (root login will be locked by security-misc)

debian/security-misc.preinst
```diff
@@ -15,6 +15,30 @@ true "
...
 
+user_name_detection() {
+   if [ "$user_name" != "" ]; then
+      return 0
+   fi
+   if [ "$USER" != "" ]; then
+      true "INFO: Setting user_name to USER $USER..."
+      user_name="$USER"
+   fi
+   if [ "$user_name" = "" ] && [ "$SUDO_USER" != "" ]; then
```
adrelanos, line 26: ` && [ "$SUDO_USER" != "root" ]`

or something.

debian/security-misc.preinst
adrelanos: Correction: Again, don't accept value of `root`.

debian/security-misc.preinst
```diff
@@ -87,12 +111,20 @@ sudo_users_check () {
...
+   ## Prevent users from locking themselves out.
+   ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/4
    if [ "$are_there_any_sudo_users" = "yes" ]; then
       return 0
    fi
 
-   ## Prevent users from locking themselves out.
-   ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/4
+   user_name_detection
+   if [ "$user_name" != "" ] && [ "$username" != "root" ]; then
```
adrelanos, line 121: [1] Instead `user_name_detection` should make sure it won't ever set `user_name` to `root`.

Also `user_name` vs `username` (missing underscore). But `[ "$username" != "root" ]` should be probably gone anyhow because of [1].

debian/security-misc.preinst
```diff
@@ -159,6 +191,14 @@ console_users_check() {
       return 0
    fi
 
+   user_name_detection
+   if [ "$user_name" != "" ] && [ "$user_name" != "root" ]; then
```
adrelanos, line 195: same as https://github.com/Whonix/security-misc/pull/74/files#r403455337

h0tw4t3r:
I decided to avoid using global variables, now it either returns non-root username, or nothing

debian/security-misc.preinst
```diff
@@ -15,6 +15,53 @@ true "
 #####################################################################
 "
 
+saved_user_name = ""
+get_user_name() {
+   if [ "$saved_user_name" != "" ]; then
+      return $saved_user_name
```
adrelanos, line 21: bash cannot return strings, only numbers.

And please quote your variables.

debian/security-misc.preinst
```diff
@@ -15,6 +15,53 @@ true "
 #####################################################################
 "
 
+saved_user_name = ""
```
adrelanos, line 18: bash  doesn't allow spaces before and after the `=` sign.

debian/security-misc.preinst
```diff
@@ -15,6 +15,53 @@ true "
...
+   if [ "$saved_user_name" != "" ]; then
+      return $saved_user_name
+   fi
+   local user_name = $saved_user_name
+
+   ## 1st attempt - Get name from $USER
+   if [ "$USER" != "" ]; then
+      true "INFO: Setting user_name to USER $USER..."
+      user_name="$USER"
+      if ["$user_name" != "root"]; then
```
adrelanos, line 29: `["$` syntax error.

adrelanos:
Stalled. Please re-submit if still applicable.

PR #75, opened 2020-09-18
Blacklist more modules (based on OpenSCAP for RHEL 8)

flawedworld:
The modules blacklisted may  be debatable, but they aren't very commonly used so I think this shouldn't impact the average user. Open to feedback.

Sourced from: https://static.open-scap.org/ssg-guides/ssg-rhel8-guide-index.html

### DIFF
--- a/etc/modprobe.d/30_security-misc.conf
+++ b/etc/modprobe.d/30_security-misc.conf
@@ -44,6 +44,14 @@ install appletalk /bin/false
 install psnap /bin/false
 install p8023 /bin/false
 install p8022 /bin/false
+install can /bin/false
+install atm /bin/false
+
+# Disable uncommon filesystems to reduce attack surface
+install cramfs /bin/false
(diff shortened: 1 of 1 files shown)


adrelanos:
https://forums.whonix.org/t/kernel-hardening/7296/445

madaidan:
There was discussion on blacklisting filesystem modules before https://forums.whonix.org/t/blacklist-more-kernel-modules-to-reduce-attack-surface/7989

Isn't `vfat` quite common?

flawedworld:
```vfat``` is used for FAT12, FAT16 and FAT32. I would say that it's mostly an issue on a UEFI system, as it has to be used for the EFI partition. For that reason it would probably be best to omit it on second thoughts.

adrelanos:
Merged and unblacklisted `vfat`. Will also remove `squashfs`as this would likely break Whonix-Host ISO.

PR #76, opened 2020-09-18
Add IPv6 sysctl options and enforce kernel.perf_event_paranoid=3

flawedworld:


adrelanos:
https://forums.whonix.org/t/kernel-hardening/7296/445

PR #98, opened 2021-01-11
Update virusforget

dantiodev:


adrelanos:
Your intentions are appreciated but hard to tell you without potentially offending you:
These pull requests are not helpful.

Please do not create 20+ separate pull request per minor change such as copyright year per file.

If I wanted to update the copyright year, I'd use a search and replace script for the whole source folder and do it with one commit. That would take less time than clicking merge for 20+ pull requests. The idea behind pull requests is to do something what probably otherwise wouldn't happen or not happen as fast. I.e. the pull request submitter is helping the pull request merger to safe some time for the shared benefit of the project.

Updating the copyright year is not a time critical task. Not legal advice but there is literally years of time before forgetting to update the copyright year would become an issue.

Could you please close these pull requests so I don't have to?

dantiodev:
> Please do not create 20+ separate pull request per minor change such as copyright year per file.

Yep, it was my mistake not to merge the commits (the client I was using instead of web GitHub for native reasons had confusing UX)

> Updating the copyright year is not a time critical task. Not legal advice but there is literally years of time before forgetting to update the copyright year would become an issue.

Never stated otherwise, just wanted to update the year to be up to date (why not?)

> Could you please close these pull requests so I don't have to?

I can.

adrelanos:
Thank you!

Should you have any other pull request for enhancements, I am happy to
look into it.

PR #100, opened 2021-02-06
hide-hardware-info: allow unrestricting selinuxfs

0xC0ncord:
On SELinux systems, the /sys/fs/selinux directory must be visible to
userspace utilities in order to function properly.


adrelanos:
https://forums.whonix.org/t/restrict-hardware-information-to-root/7329/88

adrelanos:
https://github.com/Kicksecure/security-misc/pull/108

PR #103, opened 2021-10-09
hide-hardware-info: re-enable restrictions on sysfs when using SELinux

0xC0ncord:
An oversight on my part simply re-enabled permissions to sysfs when SELinux was in effect. This PR fixes this by explicitly setting permissions on the contents of `/sys/*` and `/sys/fs*` instead of just `/sys` if SELinux is in use.

Additionally, fix the indentation so it's consistent with the rest of the script's style.

### DIFF
--- a/usr/libexec/security-misc/hide-hardware-info
+++ b/usr/libexec/security-misc/hide-hardware-info
@@ -88,6 +88,16 @@ done
 ## properly
 if [ -d /sys/fs/selinux ]; then
     if [ "${selinux}" = "1" ]; then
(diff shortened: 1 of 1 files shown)


0xC0ncord:
Any updates? I reached out to @madaidan directly and got indication he already gave the OK on this.

adrelanos:
This took longer because I was (and still am) in process of splitting Kicksecure and Whonix source code.
Thank you for your interest and pull request!

0xC0ncord:
No worries. Thanks!

PR #104, opened 2022-06-07
Fix issues found with permission-hardening on my system

ntninja:
See commit list for details.


adrelanos:
https://forums.whonix.org/t/suid-disabler-and-permission-hardener/7706/78

PR #108, opened 2022-07-05
Continue for loop if unable to change one directory's permission

Krish-sysadmin:
I ran the script hide-hardware-info as root and I got the message "chmod: changing permissions of '/sys/fs/resctrl': operation not permitted"

### DIFF
--- a/usr/libexec/security-misc/hide-hardware-info
+++ b/usr/libexec/security-misc/hide-hardware-info
@@ -93,9 +93,9 @@ if [ -d /sys/fs/selinux ]; then
     for i in /sys/* /sys/fs/*
     do
       if [ "${sysfs_whitelist}" = "1" ]; then
-        chmod o-rwx "${i}"
+        chmod o-rwx "${i}" || continue
       else
-        chmod og-rwx "${i}"
+        chmod og-rwx "${i}" || continue
       fi
     done
     chmod o+rx /sys /sys/fs /sys/fs/selinux


Krish-sysadmin:
Also I don't know enough about the /sys virtual filesystem but this script does not seem to work well on Fedora and I was unable to boot when I tested it with the selinux mode enabled in the script even with the changes that I've made in this pull request so maybe you'd be able to look into that @adrelanos  - thank you!

0xC0ncord:
> If anything, I think `|| true` seems better here than `|| continue`?

I agree. The inability to change the attributes of `/sys/fs/resctl` just needs to be ignored.

> At time of writing, there is no security-misc maintainer for [...] security-misc selinux support.

I don't mind taking care of this, but there are many components of `security-misc` that I have never tested on an SELinux system.

adrelanos:
Will use `|| true`.

Krish-sysadmin:
Happy to test on Fedora and RHEL

Krish-sysadmin:
@adrelanos By the way, would Kicksecure work on Debian Testing, Ubuntu or Debian Sid (unstable). Have these been tested? Thank you!

adrelanos:
> @adrelanos Sorry, just saw this. That's perfect. Thanks for merging it. As for helping with testing I am more than happy to .

Thanks!

Are you subscribed to this repository or should I keep writing @Krish-sysadmin for selinux related matters? (Hope I'll remember that.)

Unfortunately I doubt github has some way to get notified about a single-file change in a repository only.

Krish-sysadmin:
I'm subscribed now @adrelanos . I'm excited to see and help this project progress. While currently in some areas proprietary operating systems and the *BSDs might have an edge over most desktop linux distros in their default configurations, I am sure we can help change this situation and help make Linux more widely adopted!

adrelanos:
Btw that's the bigger vision for Kicksecure:
https://www.kicksecure.com/wiki/Main_Page#Design_and_Development_Vision

adrelanos:
> Yeah, I see that. I'd like to have Linux security be on par or better on the desktop with say macOS. We still have an edge over Windows in numerous ways but we've been losing that with the kernel getting more bloated and more security vulnerabilities in it. It's definitely worthwhile trying to fix this. There's also the fact that Linux needs to support far more hardware, architectures and use cases like server deployments, supercomputers (compared to say Windows) just to name a few that there is often preference given for performance/functionality over security.

Yes. A stronger focus by Linux on security would be desirable. Ideally, security-misc kernel settings wouldn't be required because the Linux kernel has an easy switch to flip on all the maximum security settings (still allowing to ease it where needed on top of it).

Related, this might be of interested:
https://www.kicksecure.com/wiki/Security_Roadmap

> I'd like Kicksecure "improvements" to be able to work on not only Debian but also Fedora, Ubuntu and other mainstream distributions right out of the box

Other distributions are welcome to port it (should be comparatively easy to Debian based distributions such as Ubuntu) or package it for their distributions (Fedora uses different packaging related files). In any case, it seems comparatively easy to package since there is no compiled code.

PR #109, opened 2022-07-07
Blacklist more kernel modules

raja-grewal:
Further increased quantity and type of modules blacklisted. The majority are unlikely to be used by the typical user and in cumulative represent a large attack surface reduction especially against more legacy approaches.

Open to feedback.

In summary:
- Increased protections against DMA attacks using more obscure firewire modules,
- Disabled more uncommon file systems that should not conflict with whonix requirements,
- Disabled a variety of uncommon network file systems,
- Disabled use of CD-ROM by default as the technology seems to be all but disappearing, and
- Intel CPUs might glow less brightly in the dark with some ME blocking (https://www.kernel.org/doc/html/latest/driver-api/mei/mei.html).

https://madaidans-insecurities.github.io/guides/linux-hardening.html#kasr-kernel-modules
https://forums.whonix.org/t/blacklist-more-kernel-modules-to-reduce-attack-surface/7989
https://security.stackexchange.com/questions/4098/how-to-disable-firewire-in-openbsd-linux-to-prevent-attacks-through-firewire
https://linux-audit.com/kernel-hardening-disable-and-blacklist-linux-modules/

Finally, the use of /bin/false should perhaps be replaced with /bin/true.

This does not appear to impact whonix users but users from other distributions could benefit from this increased compatibility by avoiding log errors when using this code.

https://github.com/ComplianceAsCode/content/issues/539

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,6 +149,11 @@ of multiple vulnerabilities so it is blacklisted.
 * The MSR kernel module is blacklisted to prevent CPU MSRs from being
 abused to write to arbitrary memory.
 
+* Disables a large array of uncommon file systems and network file systems that reduces the attack surface especially against legacy approaches.
+
+* Disables the use of CD-ROM devices by default.
+
+* Provides some blocking of the interface between the [Intel Management Engine (ME)](https://www.kernel.org/doc/html/latest/driver-api/mei/mei.html) and the OS.
 ### Other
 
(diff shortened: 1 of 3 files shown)


adrelanos:
Please see:
https://forums.whonix.org/t/blacklist-more-kernel-modules-to-reduce-attack-surface/7989/27

PR #110, opened 2022-07-09
Incorporated Ubuntu’s kernel module blacklists and more verbose errors

raja-grewal:
Building on the successfully merged PR [#109](https://github.com/Kicksecure/security-misc/pull/109), goal is to add a plethora of more kernel modules to be blacklisted by default based largely on the Ubuntu defaults.

https://git.launchpad.net/ubuntu/+source/kmod/tree/debian/modprobe.d?h=ubuntu/disco 
https://forums.whonix.org/t/blacklist-more-kernel-modules-to-reduce-attack-surface/7989/13
 
Overall, this blacklists their automatic loading but does NOT permanently disable them  in order to reduce the likelihood of compatibility issues. 

I am not deeply familiar with the intricacies of each of these modules and so if others in the community are more knowledgeable, maybe we could begin permanently disabling them over time just to be safe. 

Furthermore, due to intended future compatibility with ISOs, we [decided](https://forums.whonix.org/t/blacklist-more-kernel-modules-to-reduce-attack-surface/7989/30) not to disable CD-ROM by default. Instead what if we simply blacklist the modules so they are only loaded on demand?

Also as there is a strict distinction between “disable” permanently and “blacklist” automatic loading, the comments were also correspondingly updated. Perhaps we should make is difference clearer throughout the whole project?

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,14 +88,23 @@ disabled.
 
 * IOMMU is enabled to prevent DMA attacks.
 
-### Blacklisted kernel modules
+### Disables and blacklists kernel modules
 
-Certain kernel modules are blacklisted to reduce attack surface via the
+Certain kernel modules are disabled and blacklisted by default to reduce attack surface via the
 `/etc/modprobe.d/30_security-misc.conf` configuration file.
 
 * Deactivates Netfilter's connection tracking helper - this module
 increases kernel attack surface by enabling superfluous functionality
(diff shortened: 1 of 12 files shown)


raja-grewal:
https://forums.whonix.org/t/blacklist-more-kernel-modules-to-reduce-attack-surface/7989/33

PR #111, opened 2022-07-12
Increased kernel hardening at boot

raja-grewal:
Please refer to the existing discussion beginning at:
https://forums.whonix.org/t/kernel-hardening/7296/465

In summary:
- Include mitigations by default against 2 additional known hardware-level CPU vulnerabilities (L1D Flushing and MMIO Stale Data). 
- Force the kernel to panic on "oopses"
- Enforce IOMMU TLB invalidation
- Distrust the bootloader as source of entropy
- Enables randomisation of the kernel stack offset on syscall entries
- Disables slub_debug due to kernel deciding to implicitly disable kernel pointer hashing
- Explicitly highlight a few existing sysctl defaults

### DIFF
--- a/etc/default/grub.d/40_kernel_hardening.cfg
+++ b/etc/default/grub.d/40_kernel_hardening.cfg
@@ -53,3 +53,6 @@ GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX extra_latent_entropy"
 ## https://lkml.org/lkml/2020/7/16/122
 ## https://github.com/torvalds/linux/blob/fb1201aececc59990b75ef59fca93ae4aa1e1444/Documentation/admin-guide/kernel-parameters.txt#L835-L848
 GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX debugfs=off"
+
+## Force the kernel to panic on "oopses" (which may be due to false positives)
+GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX oops=panic"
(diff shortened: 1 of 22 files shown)


etc/default/grub.d/40_kernel_hardening.cfg
```diff
@@ -53,3 +53,6 @@ GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX extra_latent_entropy"
 ## https://lkml.org/lkml/2020/7/16/122
 ## https://github.com/torvalds/linux/blob/fb1201aececc59990b75ef59fca93ae4aa1e1444/Documentation/admin-guide/kernel-parameters.txt#L835-L848
 GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX debugfs=off"
+
+## Force the kernel to panic on "oopses" (which may be due to false positives)
+GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX oops=panic"
```
adrelanos, line 58: Isn't this already implemented?
See:
`/usr/libexec/security-misc/panic-on-oops`

raja-grewal: True I am considering removing the existing implementation as  I believe setting it the beginning of the boot process may be more secure?

This way we do not have to rely on the configuration files being read by a user space service. 

https://madaidans-insecurities.github.io/guides/linux-hardening.html#sysctl

adrelanos: I think there was a specific reason why it's done as currently implemented and not as kernel command line. That's a while ago since I last re-read this discussion:
https://forums.whonix.org/t/set-oops-panic-kernel-parameter-or-kernel-panic-on-oops-1-sysctl-for-better-security/7713

Could you have a look please?

PR #114, opened 2022-11-09
Add some framebuffer drivers into blacklist

raja-grewal:
These were previously commented out to test for compatibility issues.


raja-grewal:
https://forums.whonix.org/t/kernel-hardening/7296/502

PR #119, opened 2023-01-06
Cold Boot Attack Defense Fixes

friedy10:


raja-grewal:
The way I understand it, this one [recommended](https://www.kicksecure.com/wiki/Cold_Boot_Attack_Defense#Recommendations) approach to [implement](https://www.kicksecure.com/wiki/Dev/RAM_Wipe#Design) cold boot attack defense.

However, is it possible that using `kexec` may cause some issues since its use is explicitly disabled via [`sysctl kernel.kexec_load_disabled=1`](https://github.com/Kicksecure/security-misc/blob/master/etc/sysctl.d/30_security-misc.conf#L40)? 

Note as mentioned in the `README.md`  “[k]exec is disabled as it can be used to load a malicious kernel and gain arbitrary code execution in kernel mode”.

raja-grewal:
Follow up discussion:
https://forums.whonix.org/t/kernel-hardening/7296/505

adrelanos:
Fixed:
https://forums.whonix.org/t/kernel-hardening/7296/508

PR #129, opened 2023-10-22
/var options and /var/log /var/log/audit and /usr

monsieuremre:
nosuid recommended. flatpaks are by default installed on /var, so noexec would break them. adjusted accordingly

### DIFF
--- a/usr/bin/remount-secure
+++ b/usr/bin/remount-secure
@@ -178,8 +178,7 @@ _tmp() {
 
 _var() {
    mount_folder="$NEWROOT/var"
-   ## TODO: nodev? noexec?
-   intended_mount_options="nosuid"
+   intended_mount_options="nosuid,nodev"
    remount_secure "$@"
 }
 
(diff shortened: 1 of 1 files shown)


monsieuremre:
added /var /var/log /var/log/audit and /usr. everything commented out. for testing only.

usr/bin/remount-secure
```diff
@@ -188,6 +187,44 @@ _var_tmp() {
...
+#   mount_folder="$NEWROOT/var/log/audit"
+#   intended_mount_options="nosuid,nodev,noexec"
+#   remount_secure "$@"
+#}
+
+# Not tested
+#
+# Under /etc/apt/apt.conf.d/
+# If a config file is created with the following content:
+# Dpkg::Pre-Invoke {"mount -o remount,nodev,rw /usr";};
```
adrelanos, line 216: This won't work because APT needs write access to update the package lists.

adrelanos:
Anything left do here?

Merge conflict.

And looks incomplete either way.

monsieuremre:
You can close the request. Most everything seems to be incorporated in the new script anyway. This was more of a scetch which is no longer necessary.

Issue #130, opened 2023-10-22
remounting `/usr` as read-only for better security


monsieuremre:
I have tested on a debian VM the following:

*First, mount /usr as read only
```mount -o ro --bind /usr /usr```

Try creating or modify any files under /usr
```touch /usr/bin/bad-binary```
```touch: cannot touch '/usr/bin/bad-binary': Read-only file system```

So we know it is read only. Then create the file with the following content:
```
echo "
Dpkg::Pre-Invoke {"mount -o remount,nodev,rw /usr";};
Dpkg::Post-Invoke {"mount -o remount,nodev,ro /usr";};
" >> /etc/apt/apt.conf.d/security-usr
```

Try ```apt update``` and ```apt upgrade``` or anything apt. It works.
Post and pre invoke make sure /usr is remounted everytime before apt is invoked, no matter for what purpose.

This hardens against any modification in binaries or libraries in the system. For this a remount is necessary. Only root can do this.

adrelanos:
What's the rationale for this?

Non-root users cannot write to /usr anyhow. Only root can do that. However, root could also remount read-write.

Why not use live mode instead?
https://www.kicksecure.com/wiki/Live_Mode

If intending to remount /usr as read-only then why only /usr and not others such as /lib?

monsieuremre:
Hmm. Yes. Root can already remount. This seems to bring no real benefit.

adrelanos:
Other related stuff you might be interested in:

* https://www.kicksecure.com/wiki/Dev/Strong_Linux_User_Account_Isolation
* https://www.kicksecure.com/wiki/Dev/boot_modes
* https://www.kicksecure.com/wiki/Verified_Boot
* https://forums.whonix.org/t/dm-verity-for-debian-whonix/17216

adrelanos:
I wonder if this would make more sense if the untrusted root concept was implemented.

Or if there are other users other than root writing to /usr but I am not aware of any.

monsieuremre:
Probably not much. If the current implementation of remounting succeeds, there won't be much anything left to harden regarding mount options. Tho I would then be tempted to open an issue for file-permission hardening if this new remounting hook works.

adrelanos:
Alright. Closing.

PR #131, opened 2023-10-22
/usr mounting

monsieuremre:
/usr mounting. with the newly added files, usr has to have been mounted before invoking apt, because apt is going to try to remount it. so not having it mounted will cause error. it would be mounted if the section in remount-secure works.


adrelanos:
https://github.com/Kicksecure/security-misc/issues/130

PR #133, opened 2023-10-23
Fix unmount error

monsieuremre:
With the added config files, we used the --lazy option when unmounting /var, /var/tmp and /var/log. This makes the unmount operation wait until the target is not busy. No errors occur with this patch. Shutdown takes slightly longer, but unnoticable.

### DIFF
--- a/changelog.upstream
+++ b/changelog.upstream
@@ -1,3 +1,25 @@
+commit bb14a058520b13e242fea9f3022c439c4677bd1d
+Merge: 5ed2a5c 44906e8
+Author: Patrick Schleizer <adrelanos@whonix.org>
+Date:   Wed Nov 1 11:11:54 2023 -0400
+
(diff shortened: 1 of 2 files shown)


adrelanos:
From the `umount` man page.

> -l, --lazy
> 
> Lazy unmount. Detach the filesystem from the file hierarchy now, and clean up all references to this filesystem as soon as it is not busy anymore.
>
> A system reboot would be expected in near future if you’re going to use this option for network filesystem or local filesystem with submounts. The recommended use-case for umount -l is to prevent hangs on shutdown due to an unreachable network share where a normal umount will hang due to a downed server or a network partition. Remounts of the share will not be possible.

monsieuremre:
So as I stated exactly. This fixes the issue. The slight slowdown on shutdown is the only downside. I am currently trying another workaround which would not have this issue.

PR #134, opened 2023-10-23
Fix double mount issue for /var/log and /var/tmp

monsieuremre:
Mounting var with bind and mounting a subdirectory causes /var/tmp and /var/log bind mounted twice each. can be checked with lsblk. When we bind mount var only after having mounted the subdirectories, everything is mounted only one.


adrelanos:
Great!

PR #135, opened 2023-10-23
Kernel hardening fix

monsieuremre:
For some reason some of the hardening options we declare in /etc/sysctl.d/ do not come into effect. One of them is ```fs.protected_fifos=2```

After having installed security-misc, go to the terminal and run
```sudo sysctl -a | grep fifo```
You are going to see that
```fs.protected_fifos = 1```

To fic this, remove the trailing 30_ from files. With this modification, we always get ```fs.protected_fifos = 2```. I do not know the explanation but the reason seems to be the ordering of the files. If files come alphabetically after the 99-sysctl.conf file link, then some of them get overridden with the default debian values. I don't know why this is, I don't know how this is. All I know is that changing the file names to come after this link fixes the issue.

We should also set kernel.core_uses_pid=1. Which is not in the scope of this pull.

### CODE REVIEW

**Comment by adrelanos** [comment]
Good catch. This is weird.

If we do this, we also need do update this file to delete the obsoleted config files:
https://github.com/Kicksecure/security-misc/blob/master/debian/security-misc.maintscript

Survey...

    find /etc/sysctl.d    

```
/etc/sysctl.d
/etc/sysctl.d/80-qubes.conf
/etc/sysctl.d/30_silent-kernel-printk.conf
/etc/sysctl.d/99-sysctl.conf
/etc/sysctl.d/30_security-misc_aslr-mmap.conf
/etc/sysctl.d/20_tcp_timestamps.conf
/etc/sysctl.d/README.sysctl
/etc/sysctl.d/81-qubes.conf.optional
/etc/sysctl.d/30_security-misc_kexec-disable.conf
/etc/sysctl.d/30-lkrg-virtualbox.conf
/etc/sysctl.d/30_security-misc.conf
```

Maybe the `_` is breaking it and maybe the `-` would work? In that case we might need to get rid of `_`.

**Comment by monsieuremre** [comment]
I am going to try this and report. But my understanding is, it is the order of the files. Starting with any number seems to be the problem.

**Comment by monsieuremre** [comment]
Tested. It has nothing to do with ```-``` or ```_```. All that matters is the first character. With ```-``` instead of ```_``` or instead nothing in between, all fail. If the first character is 3, or another number, it fails. Thats all that matters. So this fix is appropriate indeed.

**Comment by adrelanos** [comment]
This is super weird and not what `/etc/sysctl.d/README.sysctl` is saying. This also needs an upstream bug report.

I'd like to keep some sort of numbers because that's the convention. The actual number doesn't matter much as long as it makes sense. 30 or 40 wouldn't make much a difference for the prefix.

As per man `sysctl.d` there are many folders:

       /etc/sysctl.d/*.conf
       /run/sysctl.d/*.conf
       /usr/local/lib/sysctl.d/*.conf
       /usr/lib/sysctl.d/*.conf
       /lib/sysctl.d/*.conf
       /etc/sysctl.conf

As package we might as per convention not be supposed to use /etc anyhow. After identifying the appropriate folder for a package (as per FHS) we could check if that works better.

**Comment by adrelanos** [comment]
> Packages should install their configuration files in /usr/lib/ (distribution packages)

    /usr/lib/sysctl.d/*.conf

> Files in /etc/ are reserved for the local administrator,

> It is recommended to prefix all filenames with a two-digit number and a dash, to simplify the ordering of the files.

**Comment by adrelanos** [comment]
Rather than expecting that sysctl or systemd messed up something as simple as parsing in lexical order, where there should have been other people having had this issue and reported it before... How about there's an existing configuration file by a different package which messes with our desired setting? :)

    dpkg -S 99-protect-links.conf

> procps: /usr/lib/sysctl.d/99-protect-links.conf

    cat /usr/lib/sysctl.d/99-protect-links.conf

> fs.protected_fifos = 1

That is progress. Now we just need to find a filename which can overrule that. Probably `100-..` or higher would do. So let's go from `30` to `130`?

**Comment by adrelanos** [comment]
Fixed in git. New file location:

`/usr/lib/sysctl.d/990-security-misc.conf`

**Comment by adrelanos** [comment]
Confirmed fixed after package upgrade.

    sudo sysctl -a | grep fifo

> fs.protected_fifos = 2

**Comment by adrelanos** [comment]
To check the order in which files are parsed:

    sudo sysctl -p --system

```
* Applying /etc/sysctl.d/20_tcp_timestamps.conf ...
* Applying /etc/sysctl.d/30-lkrg-virtualbox.conf ...
* Applying /usr/lib/sysctl.d/30-qubes-gui-agent.conf ...
* Applying /usr/lib/sysctl.d/30-tracker.conf ...
* Applying /etc/sysctl.d/30_security-misc_aslr-mmap.conf ...
* Applying /usr/lib/sysctl.d/30_security-misc_kexec-disable.conf ...
* Applying /usr/lib/sysctl.d/30_silent-kernel-printk.conf ...
* Applying /etc/sysctl.d/40_debug-misc.conf ...
* Applying /usr/lib/sysctl.d/50-bubblewrap.conf ...
* Applying /usr/lib/sysctl.d/50-pid-max.conf ...
* Applying /etc/sysctl.d/80-qubes.conf ...
* Applying /usr/lib/sysctl.d/99-protect-links.conf ...
* Applying /etc/sysctl.d/99-sysctl.conf ...
* Applying /usr/lib/sysctl.d/990-security-misc.conf ...
* Applying /etc/sysctl.conf ...
* ...
```

Looks good because `/usr/lib/sysctl.d/99-protect-links.conf ` is parsed first and `/usr/lib/sysctl.d/990-security-misc.conf` is parsed afterwards.

Issue #132, opened 2023-10-23
Hardening file permissions


monsieuremre:
Hardening file permissions and removing suid bits from binaries are quite different tasks. Handling them in the same service is not optimal because on of the two is significantly simple and breaks nothing. That is file permissions. These two functionalities have to be seperated.

Manually removing suid bits essentially is not optimal because we would have to maintain a long whitelist. This has no place here. This can be implemented in the apparmor-profiles-everything repo by removing the set_suid capability from the init profile and having seperate profiles for all suid applications. These can be ported from https://github.com/roddhjav/apparmor.d.

When it comes to file permission hardneing, the following hardening options should be applied upon install and be reverted on uninstall:
```
    # Restrict access to compilers from user account to prevent creation of malicious binaries
    # Even if they are not there. When they are installed in the future, these will apply, because of the override.
    dpkg-statoverride --add --update root root 0700 /usr/bin/as
    dpkg-statoverride --add --update root root 0700 /usr/bin/g++
    dpkg-statoverride --add --update root root 0700 /usr/bin/gcc
    dpkg-statoverride --add --update root root 0700 /usr/bin/cc
    dpkg-statoverride --add --update root root 0700 /usr/bin/clang

    # Protect config files and cron
    dpkg-statoverride --add --update root root 0400 /etc/cups/cupsd.conf
    dpkg-statoverride --add --update root root 0600 /boot/grub/grub.cfg
    dpkg-statoverride --add --update root root 0600 /etc/syslog.conf
    dpkg-statoverride --add --update root root 0700 /root/.ssh
    dpkg-statoverride --add --update root root 0600 /etc/crontab
    dpkg-statoverride --add --update root root 0700 /etc/cron.d
    dpkg-statoverride --add --update root root 0700 /etc/cron.daily
    dpkg-statoverride --add --update root root 0700 /etc/sudoers.d
    dpkg-statoverride --add --update root root 0700 /etc/cron.hourly
    dpkg-statoverride --add --update root root 0700 /etc/cron.weekly
    dpkg-statoverride --add --update root root 0700 /etc/cron.monthly

    # Make sure CISOfy recommended options are applied to the following
    dpkg-statoverride --add --update root root 0644 /etc/group
    dpkg-statoverride --add --update root root 0644 /etc/group-
    dpkg-statoverride --add --update root root 0644 /etc/hosts.allow
    dpkg-statoverride --add --update root root 0644 /etc/hosts.deny
[...]

PR #139, opened 2023-10-26
Restrict access to compilers

monsieuremre:
This can also be implemented as a script pair. ```lock_compiler``` and  ```lock_compiler_undo``` with only root having access to these. Then it might be easier to use compilers without root when needed. But this does not remove functionality anyway. We sould just have to ```sudo gcc```.

### DIFF
--- a/etc/permission-hardening.d/30_compilers.conf
+++ b/etc/permission-hardening.d/30_compilers.conf
@@ -1,0 +1,5 @@
+/usr/bin/as 0700 root root
+/usr/bin/g++ 0700 root root
+/usr/bin/gcc 0700 root root
+/usr/bin/cc 0700 root root
+/usr/bin/clang 0700 root root


adrelanos:
About `sudo gcc`, please see: https://github.com/Kicksecure/security-misc/issues/132#issuecomment-1777587637

monsieuremre:
I see the point. At this time, we can just disable compilers all together. That is of course a real pain for usability especially for devs. Or this can just be an opt-in feature, using a boot parameter or something else.

adrelanos:
design discussion on how to enable/disable this latest draft and discussion here:
https://github.com/Kicksecure/security-misc/issues/132#issuecomment-1783873459

adrelanos:
As said in the other ticket:

So we don't have to parse on/off for each, best to make a syntax similar to systemctl. Here:

* `permission-hardener enable all`
* `permission-hardener disable all`
* `permission-hardener enable compiler`
* `permission-hardener disable compiler`
* `permission-hardener enable interpreter`
* `permission-hardener disable interpreter`

adrelanos:
> Hmm. To be honest this is no piece of cake to implement.

It's not that difficult.

Currently we got:

* permission-hardener
* permission-hardener-undo

Overloading permission-hardener with the config management would be difficult indeed. So first step would be renaming permission-hardener to permission-hardener-enable or so.

Then the logic of config management could be handled by a new, yet to be invented script.

I might get there one day but not a priority right now.

> This is actually extremely simple to implement in a full system apparmor policy. Want to run compilers -> gotta use sudo.

As I said before:

> Not great. Better to re-enable compiler access than running compilers with root. Because that leads to follow-up issues. Maybe source code refusing to compile or failing to compile (missing environment variables, different `PATH`, different user, missing configs/files) and confusing resulting file permissions.

monsieuremre:
We might be better off just not touching the compilers and recommending users to uninstall them if they don't do development. This enable-disable solution looks a little too much at the users end. This package would be better if it is just a install and forget kind of thing.

adrelanos:
monsieuremre:
> We might be better off just not touching the compilers and recommending users to uninstall them if they don't do development. 

This won't work. Sometimes compilers or interpreters are dependencies of 
other packages.

 > This enable-disable solution looks a little too much at the users end.
 > This package would be better if it is just a install and forget kind 
of thing.

Yes. The scope of security-misc still isn't well defined and debatable. 
Keeping security-misc still useful outside of Kicksecure, Whonix might 
mandate not disabling compilers/interpreters by default.

However, most of the code could be in security-misc and then other 
packages could have a config snippet or run a command to enable these 
protections (compilers lock, interpreter lock).

monsieuremre:
I don't find this approach to be productive. Hardening as in 'secure by default' should require no further user interaction, let alone forcing them to run commands with options. So for the time being, we can let the compilers be. I'm closing the issue.

adrelanos:
Compiler lock still planned. Ticket here:
https://phabricator.whonix.org/T941

But this will take time.

(
Ticket will be migrated to the forums one day.
https://forums.whonix.org/t/abolishing-whonix-phabricator-issue-tracker-moving-issue-tracking-to-forums-migrating-phabricator-whonix-org-to-forums-whonix-org/7112
)

PR #140, opened 2023-10-26
New lines in default permission config

monsieuremre:
These are the recommended security settings in lynis.

### DIFF
--- a/etc/permission-hardening.d/30_default.conf
+++ b/etc/permission-hardening.d/30_default.conf
@@ -67,12 +67,32 @@
 ######################################################################
 
 /home/ 0755 root root
-/home/user/ 0700 user user
 /root/ 0700 root root
 /boot/ 0700 root root
 /etc/permission-hardening.d 0600 root root
 /usr/local/etc/permission-hardening.d 0600 root root
 /lib/modules/ 0700 root root
+/usr/src 0700 root root
+/etc/cups/cupsd.conf 0400 root root
+/boot/grub/grub.cfg 0600 root root
+/etc/syslog.conf 0600 root root
+/etc/crontab 0600 root root
+/root/.ssh 0700 root root
+/etc/cron.d 0700 root root
+/etc/cron.daily 0700 root root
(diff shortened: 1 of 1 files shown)


etc/permission-hardening.d/30_default.conf
```diff
@@ -67,12 +67,32 @@
...
 /boot/ 0700 root root
 /etc/permission-hardening.d 0600 root root
 /usr/local/etc/permission-hardening.d 0600 root root
 /lib/modules/ 0700 root root
+/usr/src 0700 root root
+/etc/cups/cupsd.conf 0400 root root
+/boot/grub/grub.cfg 0600 root root
+/etc/syslog.conf 0600 root root
+/etc/crontab 0600 root root
+/root/.ssh 0700 root root
```
adrelanos, line 80: /root/.ssh is superfluous because already hardening /root in line 70?

etc/permission-hardening.d/30_default.conf
```diff
@@ -67,12 +67,32 @@
...
 /home/ 0755 root root
-/home/user/ 0700 user user
 /root/ 0700 root root
 /boot/ 0700 root root
 /etc/permission-hardening.d 0600 root root
 /usr/local/etc/permission-hardening.d 0600 root root
 /lib/modules/ 0700 root root
+/usr/src 0700 root root
+/etc/cups/cupsd.conf 0400 root root
+/boot/grub/grub.cfg 0600 root root
```
adrelanos, line 77: /boot/grub/grub.cfg is superfluous because already hardening /boot/ in line 71?

etc/permission-hardening.d/30_default.conf
monsieuremre: yes sir. it is a defense in depth and is harmless. but we can get rid of it.

monsieuremre:
These are all just defense in depth mechanisms. They are harmless. If we don't want them I can get rid of them.

adrelanos:
Merged and duplicates removed.

Great stuff! Thank you!

PR #141, opened 2023-10-26
New permission-lockdown

monsieuremre:
Cleaner and better, more functionaltiy and less prone to errors.

### DIFF
--- a/usr/libexec/security-misc/permission-lockdown
+++ b/usr/libexec/security-misc/permission-lockdown
@@ -32,35 +32,28 @@
 # /usr/libexec/security-misc/permission-lockdown: user: geoclue | chmod o-rwx "/var/lib/geoclue"
 
 home_folder_access_rights_lockdown() {
-   shopt -s nullglob
+   # Each users home directory to himself
+   for user in $(dir /home); do # lists directories only
+      if [ grep -q "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
+         if [ -f /var/cache/security-misc/state-files/$user ]
+            continue # only doing once
+         fi
+         touch "/var/cache/security-misc/state-files/$user # so that we know we did this one
+         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
+         find /home/$user -type d | while read directory; # doing this to all directories in the users home
+         do
+            chown $user $directory
+            chmod 700 $directory
+         done
 
-   ## Not using dotglob.
-   ## touch /var/cache/security-misc/state-files//home/.Trash
-   ## touch: cannot touch '/var/cache/security-misc/state-files//home/.Trash': No such file or directory
-
-   local folder_name base_name
-
-   for folder_name in /home/* ; do
-      base_name="$(basename "$folder_name")"
-      if [ -f "/var/cache/security-misc/state-files/$base_name" ]; then
-         continue
+# not doing it because this sets all files to non executable
+#         find /home/$user -type f | while read file; # doing this to all files in the users home
+#         do
+#            chown $user $file
+#            chmod 600 $file
+#         done
       fi
-      if [ ! -d "$folder_name" ]; then
-         continue
-      fi
-      if [ "$folder_name" = "/home/" ]; then
-         continue
-      fi
-      mkdir -p /var/cache/security-misc/state-files
-      echo "$0: chmod o-rwx \"$folder_name\""
-      chmod o-rwx "$folder_name"
-      ## Create a state-file so we do this only once.
-      ## Therefore a user who will manually undo this, will not get
-      ## annoyed by this being done over and over again.
-      touch "/var/cache/security-misc/state-files/$base_name"
    done
-
-   shopt -u nullglob
 }
 
 home_folder_access_rights_lockdown
--- a/usr/libexec/security-misc/permission-lockdown
+++ b/usr/libexec/security-misc/permission-lockdown
@@ -34,9 +34,13 @@
 home_folder_access_rights_lockdown() {
    # Each users home directory to himself
    for user in $(dir /home); do # lists directories only
+      if [ -f /var/cache/security-misc/state-files/$user ]; then
+         continue
+      fi
       if [ $(id --user $user) ]; then # check if user actually exists, and this is not some random directory
-         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
+         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user
          echo "Permission updated: chmod go-rwx /home/$user"
+         touch /var/cache/security-misc/state-files/$user # so that we know we did this one
       fi
    done
 }
--- a/usr/libexec/security-misc/permission-lockdown
+++ b/usr/libexec/security-misc/permission-lockdown
@@ -34,7 +34,7 @@
 home_folder_access_rights_lockdown() {
    # Each users home directory to himself
    for user in $(dir /home); do # lists directories only
-      if [ grep -q "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
+      if [ grep --quiet "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
          dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
          echo "Permission updated: chmod go-rwx /home/$user"
       fi
--- a/usr/libexec/security-misc/permission-lockdown
+++ b/usr/libexec/security-misc/permission-lockdown
@@ -34,7 +34,7 @@
 home_folder_access_rights_lockdown() {
    # Each users home directory to himself
    for user in $(dir /home); do # lists directories only
-      if [ grep --quiet "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
+      if [ $(id --user $user) ]; then # check if user actually exists, and this is not some random directory
          dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
          echo "Permission updated: chmod go-rwx /home/$user"
       fi
--- a/usr/libexec/security-misc/permission-lockdown
+++ b/usr/libexec/security-misc/permission-lockdown
@@ -38,21 +38,8 @@ home_folder_access_rights_lockdown() {
          if [ -f /var/cache/security-misc/state-files/$user ]
             continue # only doing once
          fi
-         touch "/var/cache/security-misc/state-files/$user # so that we know we did this one
          dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
-         find /home/$user -type d | while read directory; # doing this to all directories in the users home
-         do
-            chown $user $directory
-            chmod 700 $directory
-         done
-
-# not doing it because this sets all files to non executable
-#         find /home/$user -type f | while read file; # doing this to all files in the users home
-#         do
-#            chown $user $file
-#            chmod 600 $file
-#         done
-      fi
+         touch /var/cache/security-misc/state-files/$user # so that we know we did this one
    done
 }
 
--- a/usr/libexec/security-misc/permission-lockdown
+++ b/usr/libexec/security-misc/permission-lockdown
@@ -35,11 +35,9 @@ home_folder_access_rights_lockdown() {
    # Each users home directory to himself
    for user in $(dir /home); do # lists directories only
       if [ grep -q "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
-         if [ -f /var/cache/security-misc/state-files/$user ]
-            continue # only doing once
-         fi
          dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
-         touch /var/cache/security-misc/state-files/$user # so that we know we did this one
+         echo "Permission updated: chmod go-rwx /home/$user"
+      fi
    done
 }
 


adrelanos:
* Could you use shellcheck please? It will probably suggest to quote some variables.
* `local`ize variables please.
*  `touch "/var/cache/security-misc/state-files/$user` please only after `dpkg-statoverride` because otherwise `touch`might succeed but `dpkg-statoverride` fails. Then we'd have a done file while it was actually not done.

```
         find /home/$user -type d | while read directory; # doing this to all directories in the users home
         do
            chown $user $directory
            chmod 700 $directory
         done
```

This is unnecessary. If the /home/user is off-limits so is any sub folder. Unnecessarily risky, intrusive. This could do stuff unwanted by the user.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,28 @@
...
+#         done
       fi
-      if [ ! -d "$folder_name" ]; then
-         continue
-      fi
-      if [ "$folder_name" = "/home/" ]; then
-         continue
-      fi
-      mkdir -p /var/cache/security-misc/state-files
-      echo "$0: chmod o-rwx \"$folder_name\""
```
adrelanos, line 55: Please keep user visible output to inform the user what change is being performed.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,28 @@
...
+      if [ grep -q "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
+         if [ -f /var/cache/security-misc/state-files/$user ]
+            continue # only doing once
+         fi
+         touch "/var/cache/security-misc/state-files/$user # so that we know we did this one
+         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user itself
+         find /home/$user -type d | while read directory; # doing this to all directories in the users home
+         do
+            chown $user $directory
+            chmod 700 $directory
```
adrelanos, line 46: I find `chmod o-rwx` basically saying `others` `minus` `read`, `write`, `execute` more descriptive than an obscure `700` that I need to look up what it does. Should be effectively the same.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,28 @@
 # /usr/libexec/security-misc/permission-lockdown: user: geoclue | chmod o-rwx "/var/lib/geoclue"
 
 home_folder_access_rights_lockdown() {
-   shopt -s nullglob
```
adrelanos, line 35: Getting ride of `shopt` is nice.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,28 @@
...
-
-   local folder_name base_name
-
-   for folder_name in /home/* ; do
-      base_name="$(basename "$folder_name")"
-      if [ -f "/var/cache/security-misc/state-files/$base_name" ]; then
-         continue
+   # Each users home directory to himself
+   for user in $(dir /home); do # lists directories only
+      if [ grep -q "$user" /etc/passwd ]; then # check if user actually exists, and this is not some random directory
```
adrelanos, line 37: * Need to drop the `[`, `]`?
* `--quiet` please.
* Reading `/etc/passwd` is best avoided because then this will show up for apparmor-profile-everything.
* Is this really needed?

Do we really need to avoid "random directories"? I think covering these random directories is quite important irrespective of `/etc/passwd`. This is because the user might have restored a backup but not yet re-created the user.

For example, in Qubes App Qubes I sometimes do `sudo adduser user2` so I can then experiment under user `user2`. That creates a home folder. I personally don't put interesting information there. It's just for testing. But others might have other use cases.

After reboot of the App Qube, user `user2` will be done because Qubes does not persist the root image for App Qubes by default. The new home folder `/home/user2` however still exists. That would then be a "random folder".

If security-misc is installed in such or a similar situation, it would be best to cover these folders. And this is only done once. Then the lockfile prevents doing this again.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,28 @@
...
-   shopt -s nullglob
-
-   ## Not using dotglob.
-   ## touch /var/cache/security-misc/state-files//home/.Trash
-   ## touch: cannot touch '/var/cache/security-misc/state-files//home/.Trash': No such file or directory
-
-   local folder_name base_name
-
-   for folder_name in /home/* ; do
-      base_name="$(basename "$folder_name")"
```
adrelanos, line 44: Getting rid of basename is also nice.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,28 @@
...
 home_folder_access_rights_lockdown() {
-   shopt -s nullglob
-
-   ## Not using dotglob.
-   ## touch /var/cache/security-misc/state-files//home/.Trash
-   ## touch: cannot touch '/var/cache/security-misc/state-files//home/.Trash': No such file or directory
-
-   local folder_name base_name
-
-   for folder_name in /home/* ; do
```
adrelanos, line 43: The `*` is best avoided indeed.

monsieuremre:
> * Could you use shellcheck please? It will probably suggest to quote some variables.
> 
>     * `local`ize variables please.
> 
>     * `touch "/var/cache/security-misc/state-files/$user` please only after `dpkg-statoverride` because otherwise `touch`might succeed but `dpkg-statoverride` fails. Then we'd have a done file while it was actually not done.
> 
> 
> ```
>          find /home/$user -type d | while read directory; # doing this to all directories in the users home
>          do
>             chown $user $directory
>             chmod 700 $directory
>          done
> ```
> 
> This is unnecessary. If the /home/user is off-limits so is any sub folder. Unnecessarily risky, intrusive. This could do stuff unwanted by the user.

Yes. The only reason I did this is because, well, the old implementation did it. Normally we can just very easily do just the home directories.

usr/libexec/security-misc/permission-lockdown
monsieuremre: Yes, readability. I can change this. But this isnt o-rwx. It is go-rwx. Only the user has permissions, no others and no one in the group.

monsieuremre:
How does it look now?

usr/libexec/security-misc/permission-lockdown
adrelanos: What's the issue with randomly created directories in /home?

monsieuremre: There is a directory ```/home/foo```. We try making the user ```foo``` own this directory. The user does not exist.

usr/libexec/security-misc/permission-lockdown
adrelanos: The `g`  for `group` isn't needed.
Debian by default uses `USERGROUPS=yes` in `/etc/adduser.conf`. The group which the user is being added to has the same name as the user. If the username is `user` then the name of the group is also `user`.
This could use an explanatory comment as this isn't well known nowadays. Information is scattered on `debian.org` etc.

Some background information here:
https://unix.stackexchange.com/questions/156473/reasons-behind-the-default-groups-and-users-on-linux

In short, this is useful for "file sharing". A if `user1` wants to share data with `user2` the command required to run is `sudo addgroup user1 user2`.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,17 @@
...
+   # Each users home directory to himself
+   for user in $(dir /home); do # lists directories only
+      if [ -f /var/cache/security-misc/state-files/$user ]; then
          continue
       fi
-      if [ ! -d "$folder_name" ]; then
-         continue
+      if [ $(id --user $user) ]; then # check if user actually exists, and this is not some random directory
+         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user
+         echo "Permission updated: chmod go-rwx /home/$user"
```
adrelanos, line 42: The echo before dpkg-statoverride please. This is because should dpkg-statoverride hang (bugs, database locked) then it's good if clues are given in the terminal to avoid a mysterious issue trying to figure out why it's hanging.

Also if running dpkg-statoverride we should echo exactly that. The user gets informed what is actually done. The actual command.

Could argue if the command is too obscure then a different INFO message would be better.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,17 @@
...
-      if [ -f "/var/cache/security-misc/state-files/$base_name" ]; then
+   # Each users home directory to himself
+   for user in $(dir /home); do # lists directories only
+      if [ -f /var/cache/security-misc/state-files/$user ]; then
          continue
       fi
-      if [ ! -d "$folder_name" ]; then
-         continue
+      if [ $(id --user $user) ]; then # check if user actually exists, and this is not some random directory
+         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user
```
adrelanos, line 41: Now that we are reconsidering this... Did you test if dpkg-statoverride works for files not managed by dpkg, i.e. folders in the home folder?

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,17 @@
...
-      base_name="$(basename "$folder_name")"
-      if [ -f "/var/cache/security-misc/state-files/$base_name" ]; then
+   # Each users home directory to himself
+   for user in $(dir /home); do # lists directories only
+      if [ -f /var/cache/security-misc/state-files/$user ]; then
          continue
       fi
-      if [ ! -d "$folder_name" ]; then
-         continue
+      if [ $(id --user $user) ]; then # check if user actually exists, and this is not some random directory
```
adrelanos, line 40: `id` also best avoided but already got another thread for that as I don't see the issue with the random directories and think it is a feature, not a bug.

usr/libexec/security-misc/permission-lockdown
monsieuremre: Yes it does. It might seem a little unnecessary tho since dpkg is not expected to touch these folders. But it can override and it can update, and it does.

usr/libexec/security-misc/permission-lockdown
adrelanos: I mean, we cannot say "we do chmod" while what we really do is "dpkg-statoverride". That would be a bug.

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,17 @@
...
-   ## touch /var/cache/security-misc/state-files//home/.Trash
-   ## touch: cannot touch '/var/cache/security-misc/state-files//home/.Trash': No such file or directory
-
-   local folder_name base_name
-
-   for folder_name in /home/* ; do
-      base_name="$(basename "$folder_name")"
-      if [ -f "/var/cache/security-misc/state-files/$base_name" ]; then
+   # Each users home directory to himself
+   for user in $(dir /home); do # lists directories only
```
adrelanos, line 36: `local user`
on top please.

monsieuremre: yes

usr/libexec/security-misc/permission-lockdown
monsieuremre: should I just remove the check?

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,17 @@
...
+      if [ -f /var/cache/security-misc/state-files/$user ]; then
```
adrelanos, line 37: Needs quotes.
`"/var/cache/security-misc/state-files/$user"`

usr/libexec/security-misc/permission-lockdown
```diff
@@ -32,35 +32,17 @@
...
          continue
       fi
-      if [ ! -d "$folder_name" ]; then
-         continue
+      if [ $(id --user $user) ]; then # check if user actually exists, and this is not some random directory
+         dpkg-statoverride --add --update $user $user 0700 /home/$user # home directory of the user
+         echo "Permission updated: chmod go-rwx /home/$user"
+         touch /var/cache/security-misc/state-files/$user # so that we know we did this one
```
adrelanos, line 43: Also needs quotes.

adrelanos:
I can do the nitpick changes (quotes) myself if you like. Let me know your preference.

adrelanos:
Also the "echo before run the command" I could do myself. Easy stuff.

usr/libexec/security-misc/permission-lockdown
adrelanos: Ok, that's a problem indeed.

Maybe that's why the old implementation used `chmod` instead of `dpkg-statoverride`?

Why change from `chmod` to `dpkg-statoverride` anyhow?

usr/libexec/security-misc/permission-lockdown
adrelanos: This is pending below discussion why `dpkg-statoverride` instead of `chmod`.

usr/libexec/security-misc/permission-lockdown
adrelanos: I would not have expected this from the man page.

https://manpages.debian.org/bookworm/dpkg/dpkg-statoverride.1.en.html

> “stat overrides” are a way to tell [dpkg(1)](https://manpages.debian.org/bookworm/dpkg/dpkg.1.en.html) to use a different owner or mode for a path when a package is installed (this applies to any filesystem object that dpkg handles, including directories, devices, etc.).

"that dpkg handels"

monsieuremre:
You can do it. Quotes are just cosmetic tho in this case, they add no functionality.

usr/libexec/security-misc/permission-lockdown
adrelanos: I guess chmod only changes the permissions but not ownership (chown).

While dpkg-overwrite does both, permissions (chmod) and ownership (chown).

While folders in /home on many Linux distributions have the issue of having insecure permissions, I haven't seen any issues with wrong ownership.

Therefore the old implementation used chmod and I considered that sufficient. Let me know if that missed something.

usr/libexec/security-misc/permission-lockdown
monsieuremre: Yes. We may choose to go with any of the two. Chances are dpkg wont touch home unless the user explicitly gives a custom innstall directory or installs a very particular package.

adrelanos:
All done. Please open a new issue should you see any issues by the commit that I added on top.

PR #142, opened 2023-10-27
ssh config

monsieuremre:
very small but significant improvement, as recommended by cisofy


adrelanos:
Would it be sane to do that for the whole /etc/ssh folder? There's also 
private keys in that folder. (Though at least owned and read/write by 
root only.)

monsieuremre:
That is not possible without breaking stuff. Access to public keys has to be granted.

PR #143, opened 2023-10-27
new lines 990-security-misc.conf

monsieuremre:
added new recommended hardening settings with comments


adrelanos:
As per FHS and/or other conventions for a package / distribution /etc 
shouldn't be used (for local system admin) but at least for systemd some 
other location such as /lib or /usr/lib should be used. Could you 
research that please and move accordingly?

Not sure about NM. For that, /etc might be correct.

monsieuremre:
This does not seem to relate to the pull request, or I do not understand what you mean.

adrelanos:
True. Disregard. Wrong ticket. My mistake.

PR #147, opened 2023-11-02
Depend on libpam-tmpdir for very solid extra security

monsieuremre:
I am not quite sure if this would be the right way to add a dependency. But we should definetely depend on this package.

https://packages.debian.org/bookworm/libpam-tmpdir
>Many programs use $TMPDIR for storing temporary files. Not all of them are good at securing the permissions of those files. libpam-tmpdir sets $TMPDIR and $TMP for PAM sessions and sets the permissions quite tight. This helps system security by having an extra layer of security, making such symlink attacks and other /tmp based attacks harder or impossible.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -15,7 +15,7 @@ Rules-Requires-Root: no
 Package: security-misc
 Architecture: all
 Depends: python3, libglib2.0-bin, libpam-runtime, sudo, adduser, libcap2-bin,
- apparmor-profile-dist, helper-scripts, libpam-modules-bin,
+ apparmor-profile-dist, helper-scripts, libpam-modules-bin, libpam-tmpdir,
  secure-delete, dmsetup, ${misc:Depends}
 Replaces: tcp-timestamps-disable, anon-gpg-tweaks, swappiness-lowest
 Description: Enhances Miscellaneous Security Settings
--- a/debian/control
+++ b/debian/control
@@ -17,7 +17,6 @@ Rules-Requires-Root: no
 
 Package: security-misc
 Architecture: all
-Replaces: tcp-timestamps-disable, anon-gpg-tweaks, swappiness-lowest
 Depends: adduser,
          apparmor-profile-dist,
          dmsetup,


adrelanos:
This would be nice. However this was discussed before and there is an unresolved issue (which breaks the build process). Investigation stalled and help would be welcome. See:
https://forums.whonix.org/t/make-symlink-attacks-and-other-tmp-based-attacks-harder-or-impossible-using-libpam-tmpdir/8488

monsieuremre:
> This would be nice. However this was discussed before and there is an unresolved issue (which breaks the build process). Investigation stalled and help would be welcome. See: https://forums.whonix.org/t/make-symlink-attacks-and-other-tmp-based-attacks-harder-or-impossible-using-libpam-tmpdir/8488

This is from another time on another debian. I will test it and try to reproduce the error. It might not even be there anymore.

monsieuremre:
> This would be nice. However this was discussed before and there is an unresolved issue (which breaks the build process). Investigation stalled and help would be welcome. See: https://forums.whonix.org/t/make-symlink-attacks-and-other-tmp-based-attacks-harder-or-impossible-using-libpam-tmpdir/8488

I am unable to reproduce this issue. The steps I followed are:

* Freshly install debian on a vm
* Gnome desktop environment, everything default settings
* Boot up first time. ```apt install libpam-tmpdir```
* Reboot
* Download this git repo as zip, uncompress
* Install the required packages to build. ```dpkg-dev```, ```debhelper```, etc.
* Run ```dpkg-buildpackage -b``` in the project directory.
* Everything buils just fine. I now have a .deb in the parent directoy.
* No errors whatsoever

Whatever issue it was, it seems to be not there anymore. Please test and tell. I think we can merge.

adrelanos:
This issue happened in the past with cowbuilder which is used during the build process of Kicksecure / Whonix because that is the cleanest way to build packages and makes sure there are no missing build dependencies.

adrelanos:
And that unfortunately is still broken.

```
dpkg-deb: building package 'pbuilder-satisfydepends-dummy' in '/tmp/satisfydepends-aptitude/pbuilder-satisfydepends-dummy.deb'.
dpkg-deb: error: failed to make temporary file (control member): No such file or directory
E: pbuilder-satisfydepends failed
```

monsieuremre:
Hmm. First question, where does Kicksecure use this? The 'user' does not normally do this, am I right? And for machines that have this package installed, it is still possible to build everything using other tools.

Is there anything in particular that makes you prefer cowbuilder and not the classic functionality from dpkg-dev? Because dpkg-dev also makes sure there are no missing build dependencies. This doesn't seem like a deal-breaker to me unless I am missing an important detail.

adrelanos:
> Hmm. First question, where does Kicksecure use this? The 'user' does not normally do this, am I right?

Right. This is a developer tool. However, some security hardening breaks the build process and I cannot use it myself then this adds a lot of extra complexity if I myself have to opt-out from a feature which is otherwise enabled by default for all users. Because if I don't use it myself, then I don't experience any other issues that users would have. Resulting in surprises, difficult to debug / non-reproducible situations.

This isn't a security argument. It's maintenance argument. I need to be capable to manage the complexity, still understand the system. This is important. Other distributions often have less features or are abandoned.

Maybe [Liberté Linux](https://dee.su/liberte) has/had (some?) better security features but it doesn't help if the maintenance effort is too high so the main developer gets burnout and throwing the towel. (I dunno what happened to Liberté Linux.)

So all things I am doing I try to do in a clean way without too many exceptions, hacks, workarounds, difficult to remember things, surprises. Bugs must be reported upstream so these hopefully one day get fixed so any workarounds do not need to be carried forever.

Also this could be just a symptom of other things that break which then would only be learned through weird bugs being reported that are time-consuming, difficult to debug.

Therefore I really want to understand this bug, at least get it reported.
 
> Is there anything in particular that makes you prefer cowbuilder and not the classic functionality from dpkg-dev?

* supports cross-platform builds
* cross suite (can build trixie packages on bookworm and vice versa) which is useful to port to newer Debian versions or push fixes for older versions (so these can upgrade)
* similar to how Debian builds official packages
* clean chroot
* no influences from dotfiles, environment variables, filesystem on the build machine
* reproducible builds
* dpkg-buildpackage cannot configure the output folder. It uses `../`.
* Temporary files are created in `../` and in the root folder of the package source code, which then also need to be managed (make clean etc).

(Some features might be derived from pbuilder which is internally uses.)

> Because dpkg-dev also makes sure there are no missing build dependencies.

But not for independent re-builds by others. Packages already installed on the host operating system needed to build but undeclared under `Build-Depends:` won't be noticed by dpkg if missing. The only way to test this is to use a minimal, clean chroot.

adrelanos:
Found a workaround.

adrelanos:
`/usr/sbin/pam-tmpdir-helper` is a new SUID which needs whitelisting. Otherwise Xfce fails to start with a dbus-launch error.

adrelanos:
Added.

adrelanos:
Weird.

    sudo apparmor-info --boot

>AVC apparmor="DENIED" operation="exec" profile="/usr/bin/systemcheck" name="/usr/sbin/pam-tmpdir-helper" comm="sudo" requested_mask="x" denied_mask="x"

adrelanos:
systemcheck
```

[ERROR] [systemcheck] Virtualizer xen
xen unsupported by Whonix developers! systemcheck aborted! (qubes_detected: true)

Using Virtualizer this together with Whonix is recommended against, because it is rarely tested. [1] [2] [3] It could be made possible, but would require more Whonix contributors.
It may already work, but is highly experimental.



systemd-detect-virt output:

xen
xen
```

The `xen` line is really duplicated.

adrelanos:
Fixing the apparmor profile fixed this issue.

All hardening can cause impossible to predict issues.

adrelanos:
Another issue breaking the boot process, grml-debootstrap, grub installation:

* https://github.com/grml/grml-debootstrap/issues/232

adrelanos:
* [pbuilder: fails with cryptic message when $TMPDIR/$TEMP != /tmp (i.e. libpam-tmpdir)](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=725434;msg=45)
* `/usr/share/doc/pbuilder/examples/D10tmp`

----

* https://github.com/calamares/calamares/issues/2269

adrelanos:
* https://github.com/Kicksecure/security-misc/issues/198

adrelanos:
[Veracrypt stopped working after updating to whonix 17](https://forums.whonix.org/t/veracrypt-stopped-working-after-updating-to-whonix-17/19752)

PR #148, opened 2023-11-02
Harden the loading of new modules to the kernel after install

monsieuremre:
Normally it is recommended to have ```kernel.modules_disabled``` set to ```1```. This is the case for all hardened systems, where they pre-define all the modules to be loaded if any. This prevents the loading of new modules to the kernel. Having this set as a sysctl configuration breaks the system, because it also denies ```systemd-modules-load.service``` the ability to load the modules that are legitemately needed. I found a workaround. I created a systemd service. It triggers on start up but only after ```systemd-modules-load.service``` has already done its thing. Then it disables the loading of new modules to the kernel by setting the aforementioned kernel variable. The system is fully functional.

The only scenario where this would be somewhat inconvenient is maybe if and only if the following exact things happen: 
* a user starts up the system
* some peripheral he uses needs drivers that is in the kernel
* the drivers are kind of niche so they don't get loaded by default
* and this device is not plugged in when the system is booting
* so the drivers for that peripheral don't load
* and the user decides to use this peripheral afterwards and plugs it
* and the drivers cannot load, because we disabled it
 
But if everything is plugged in on system startup, nothing can break even in such an edge scenario, I think.

There are some stuff I am not sure of. First of all, it works as is perfectly. I put the service under ```lib/systemd/system/harden-module-loading.service``` and the executable that sets the kernel variable in ```usr/libexec/security-misc/disable-kernel-module-loading```. If another path is more desirable for the executable, we can change that. I also enable the service after install by adding the line ```systemctl enable disable-module-loading.service``` into the file ```debian/security-misc.postinst```. I am not sure if this would be right method of enabling the service. If it needs to be taken care of somewhere else, this can also be fixed.

### DIFF
--- a/debian/security-misc.postinst
+++ b/debian/security-misc.postinst
@@ -61,6 +61,8 @@ pam-auth-update --package
 /usr/libexec/security-misc/permission-lockdown
 permission_hardening
 
+systemctl enable disable-module-loading.service
+
 ## https://phabricator.whonix.org/T377
 ## Debian has no update-grub trigger yet:
 ## https://bugs.debian.org/481542
--- a/debian/security-misc.postinst
+++ b/debian/security-misc.postinst
@@ -61,8 +61,6 @@ pam-auth-update --package
 /usr/libexec/security-misc/permission-lockdown
 permission_hardening
 
-systemctl enable disable-module-loading.service
-
 ## https://phabricator.whonix.org/T377
 ## Debian has no update-grub trigger yet:
 ## https://bugs.debian.org/481542
--- a/lib/systemd/system/harden-module-loading.service
+++ b/lib/systemd/system/harden-module-loading.service
@@ -1,0 +1,13 @@
+[Unit]
+Description=Disable the loading of modules to the kernel after startup. This could be malicious.
+After=systemd-modules-load.service
+# This functionality is implemented with this and not directly in the sysctl config is
+# to allow systemd-modules-load.service to load the modules with no problem but
+# to disallow anyone else do the same after the system boots up.
+
+[Service]
+Type=oneshot
+ExecStart=/usr/libexec/security-misc/disable-kernel-module-loading
+
+[Install]
+WantedBy=sysinit.target
--- a/lib/systemd/system/harden-module-loading.service
+++ b/lib/systemd/system/harden-module-loading.service
@@ -1,6 +1,7 @@
 [Unit]
 Description=Disable the loading of modules to the kernel after startup. This could be malicious.
 After=systemd-modules-load.service
+Before=sysinit.target
 # This functionality is implemented with this and not directly in the sysctl config is
 # to allow systemd-modules-load.service to load the modules with no problem but
 # to disallow anyone else do the same after the system boots up.
--- a/usr/libexec/security-misc/disable-kernel-module-loading
+++ b/usr/libexec/security-misc/disable-kernel-module-loading
@@ -1,0 +1,5 @@
+#!/bin/bash
+
+sysctl -w kernel.modules_disabled=1
+
+echo "The loading of new modules to the kernel has been disabled by security-misc" >&2
--- a/usr/libexec/security-misc/disable-kernel-module-loading
+++ b/usr/libexec/security-misc/disable-kernel-module-loading
@@ -2,4 +2,4 @@
 
 sysctl -w kernel.modules_disabled=1
 
-echo "The loading of new modules to the kernel has been disabled by security-misc" >&2
+echo "The loading of new modules to the kernel has been disabled by security-misc"


adrelanos:
Not sure yet. I will add some comments about the implementation now before I comment on the actual functionality.

debian/security-misc.postinst
```diff
@@ -61,6 +61,8 @@ pam-auth-update --package
 /usr/libexec/security-misc/permission-lockdown
 permission_hardening
 
+systemctl enable disable-module-loading.service
```
adrelanos, line 64: This fortunately isn't needed. `#DEBHELPER` automagically does that, it honors systemd preset and whatnot.

lib/systemd/system/harden-module-loading.service
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Disable the loading of modules to the kernel after startup. This could be malicious.
+After=systemd-modules-load.service
```
adrelanos, line 3: Below that line it use some additional `Before=`. Reasons:

* avoid race conditions
* have a more predictable chain of execution of systemd units during bootup
  * i.e. non-parallelized (or less parallelization) execution by systemd
  * avoid wondering someone in the future "why does this sometimes work and sometimes not"
* Better security. If following the rationale of "something resulting in later loading a dangerous kernel module" then this should happen only exactly `After=systemd-modules-load.service` but certainly before the desktop is fully load. Realistically the desktop won't be load that quick but we don't really now what exactly happens because the boot process uses parallelization and is therefore hard to understand.
  * The idea is to conceptually distinguish the boot process into two different phases:
    * An earlier more trusted phase where kernel modules can still be load where the probability is lower that unexpected/dangerous kernel modules are loaded.
    * A later less trusted phase where more code is loaded, when an untrusted root or malware might attempt to get a kernel module loaded.

usr/libexec/security-misc/disable-kernel-module-loading
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
```
adrelanos, line 2: Could use `set -e`. Maybe unlikely to fail but who knows what changes in the future so it is best practice to have good error handling.

adrelanos:
Your new implementation solution is is a clever approach.

Restricting kernel module loading was attempted in the past:

* https://www.kicksecure.com/wiki/Operating_System_Hardening#Module_Loading
* https://forums.whonix.org/t/enforce-kernel-module-software-signature-verification-module-signing-disallow-kernel-module-loading-by-default/7880

It stalled. There was some breakage in the past.

Is this tested? Would these things still be broken with the new implementation?

debian/security-misc.postinst
monsieuremre: Good to know.

monsieuremre:
Yes I have tested most anything. Anything under modules-load.d loads perfectly. No issues there. So this kind of breakage is not present.

When it comes to virtualbox, I think that might still break, but probably not. Because far as I know virtualbox loads its own module to the kernel for virtualization. Thats why it is also not compatible with secure boot without adjusting. But the question is, when does virtual box load its module. If it happens during boot automatically, then no breakage. Just install. It won't work, sure. Then reboot. And it will work. If virtualbox tries to load its modules only when you launch the application, then it won't work.

To be honest, there are better virtualization tools, so breaking virtualbox wouldn't be a big deal for me. To be even more honest, I think there is a good chance it doesn't break anyway because it loads the modules on startup.

Ok quick research results: for example on arch, virtualbox loads its modules on startup automatically. So it wouldn't break. They do it with the file ```/usr/lib/modules-load.d/virtualbox-host-modules-arch.conf```. If an avid virtualbox user wants to continue using it in this hardened system, he can just use the same config file in that path. (Maybe it isn't even needed because they also get loaded automatically on startup on debian too. In that case, no problem.)

monsieuremre:
On a somewhat unrelated topic, I would prefer the modules signed by Debian organization and not the user. Firstly, Debian signs everything they ship anyway. So enrolling the key would be a piece of cake. And also constantly signing stuff on the user end is not the easiest thing to maintain. Because the signing process would have to extremely restricted with a mac policy. Else, once compromised, nothing would stop the intruder from signing their stuff with the private keys that are on the users machine.

I would say this is a little unrelated tho. This pull is about loading modules at all after boot, not that if they are signed or not.

monsieuremre:
Or if you want to add some third party modules that are not signed by Debian, Kicksecure/Whonix can take it upon itself to sign everything on their end with their keys. Then the user would enroll Whonix's keys.

lib/systemd/system/harden-module-loading.service
adrelanos: Meant to say (will edit above):
Below `After=` it could use some additional `Before=`.

monsieuremre: Maybe before the login manager and any network daemons. Does it sound good?

adrelanos: This is the crux. There was "no answer" to my question how to do that.

https://lists.freedesktop.org/archives/systemd-devel/2019-December/043845.html

But maybe a new systemd target could be invented one that runs after sysinit.target but before basic.target.

monsieuremre: Actually, nevermind.  I added Before=sysinit.target. And just expected, it works. So it almost right after we load the modules, we disable loading. Before sysinit.target. This one works fortunately.

monsieuremre:
Yessir, we do it before sysinit.target now. I think that has to be enough.

lib/systemd/system/harden-module-loading.service
adrelanos: That's nice!

(My previous comment was actually in the wrong PR.)

adrelanos:
Created

* https://github.com/Kicksecure/security-misc/issues/156
 
for it.

monsieuremre:
Is there any particular issue that is preventing this from being merged, like a lack of testing of a certain application like virtualbox? Or lack of testing in general.

adrelanos:
I was adjusting the Kicksecure / Whonix VirtualBox installer today to cope up with modules load disabled or module signature verification enforcement.

https://forums.whonix.org/t/whonix-linux-installer-development-discussion/15917/147
https://github.com/Kicksecure/usability-misc/commits/master

adrelanos:
Also testing this isn't simple. Specifically testing this with Qubes VM kernel is difficult because that itself is difficult to use. Trying to resolve 1 issue and running into another issue.

usr/libexec/security-misc/disable-kernel-module-loading
adrelanos: Done.

PR #151, opened 2023-11-02
Add more hardening options to the default profile in etc

monsieuremre:
Disabling coredumps and setting a secure umask value. Comments explain more in detail. I am really sorry for the many pull requests, this is the last one. I don't mean to spam the project. They are mostly really small but significant changes. I thought that it would be easier to manage for the maintainer if they are seperated.

### DIFF
--- a/etc/profile.d/30_security-misc.sh
+++ b/etc/profile.d/30_security-misc.sh
@@ -14,3 +14,11 @@ if [ -x /usr/libexec/security-misc/panic-on-oops ]; then
    ## Hide output. Otherwise could confuse Qubes UpdatesProxy.
    sudo --non-interactive /usr/libexec/security-misc/panic-on-oops 1>/dev/null 2>/dev/null
 fi
+
+# Set the default umask 0077. Whatever the access value the umask value is subtracted from it. Subtracting 7 means stripping
+# all capabilities. This means that new files and folders created by a user will have no permission for group and others. 
+umask 0077
+
+## Once again disable core dumps by limiting the coredump file size to 0.
+## One of the many places we disable coredumps in this package.
+ulimit -c 0


adrelanos:
The group does not need to be restricted. If username is `user2` then the group on Debian is `user2`, which is fine. Quote from permission-lockdown script:

```
      ## Debian by default uses USERGROUPS=yes in /etc/adduser.conf.
      ## The group which the user is being added to has the same name as the user.
      ## If the username is user then the name of the group is also user.
      ## Some background information here:
      ## https://unix.stackexchange.com/questions/156473/reasons-behind-the-default-groups-and-users-on-linux
      ## In short, this is useful for "file sharing". A if user1 wants to share data with user2 the command
      ## required to run is sudo addgroup user1 user2.
```

----

In the past, we've changed the default umask (though through a different implementation), had run into unreasonable issues and then settled for a different solution (permission-lockdown). That's quite some time ago and I'd have to re-read this:
https://forums.whonix.org/t/change-default-umask/7416

----

Changes in `/etc/profile.d` can result in difficult to predict, strange, hard to debug issues such as this:

* https://github.com/QubesOS/qubes-issues/issues/8610

where some extra output by `/etc/profile.d` resulted in broken upgrades.

----

The suggested umask change:
Really needed? I mean... Is there any example where a created file currently has insecure permissions thereby making it readable by another Linux user account?

----

The suggested `ulimit -c 0` to disable core dumps change:
Really needed? How can core dumps still be created without that setting?

adrelanos:
All your contributions are most helpful and appreciated!

monsieuremre:
> Really needed? How can core dumps still be created without that setting?
No. They can't be created. But disabling core dumps with all means possible is a better approach to not let any possibility of the user accidentally enabling them.

>  Really needed? I mean... Is there any example where a created file currently has insecure permissions thereby making it readable by another Linux user account?

Debian uses 0022 per default. It means when we create a directory for example, the permission is set to 755. This means, the group and others can read and execute, but they can't write. With 0077 we would deny them even reading. All the security guides I read recommend a minimum of 0027 umask. Debian's official [guide](https://www.debian.org/doc/manuals/securing-debian-manual/ch04s11.en.html#id-1.5.14.19) is one of them.

I know this does not break anything on a personal computer, or on a vm. But I don't know anything about qubes. They do their stuff differently and there are more things to consider there.

monsieuremre:
Having read the old forum threads, I gather that most problems that are not qubes related are because you chose to not set the group to 7 as well. A umask of 0077 would eliminate these problems. When it comes to qubes, I don't know what the problem was, and I can't find what the problem was in detail. So as I said, we are most likely clear apart from qubes. But I don't know if the problem with qubes was the umask value itself or the fact that we put files under ```/etc/profile.d/``` at all. Are there more resources?

adrelanos:
monsieuremre:
>> Really needed? How can core dumps still be created without that setting?
> No. They can't be created. But disabling core dumps with all means possible is a better approach to not let any possibility of the user accidentally enabling them.

How would core dumps we accidentally enabled?

And then how would setting this prevent this issue?

adrelanos:
Why is `ulimit -c 0` extra, duplicated? If doing `ulimit -c 0` then by extension would we have to duplicate all the sysctl etc.?

adrelanos:
As for default umask, could we use libpam-umask package instead as Debian mentions in the link that you shared? That seems cleaner.

/etc/profile.d seems hackish. Last resort only. It's hard to know in which situations it applies and when it doesn't in the various environments such as:

* different shells
* terminal emulator
* X
* Wayland
* login shell
* terminal emulator
* non-interactive session

Check this out:
https://github.com/Kicksecure/security-misc/tree/master/usr/share/pam-configs

adrelanos:
https://github.com/Kicksecure/security-misc/blob/master/usr/share/pam-configs/mkhomedir-security-misc uses `027`, which should be fine as per https://github.com/Kicksecure/security-misc/pull/151#issuecomment-1791160187.

adrelanos:
> As for default umask, could we use libpam-umask package instead as Debian mentions in the link that you shared? That seems cleaner.

https://github.com/Kicksecure/security-misc/blob/master/usr/share/pam-configs/umask-security-misc

adrelanos:
As for umask, that seems to work. Please test and let me know if that resolves the umask part.

adrelanos:
That umask setting works for user and root but it has the [same issue](https://forums.whonix.org/t/change-default-umask/7416/53) as before.

Files created by root in /etc aren't readable by "others".

> Now tb-starter package has this issue:
>
> bash -n /etc/torbrowser.d/50_user.conf
> bash: /etc/torbrowser.d/50_user.conf: Permission denied

Maybe `update-torbrowser` which runs as user `user` reading config files in `/etc` is a weird implementation and therefore should get a special fix and call it a day? If this was the case, then yes. But I don't think so. Others applications implemented in the same way:

* `/etc/thunderbird`
* `/etc/firefox-esr`

Therefore I reverted my pam based implementation for now.

adrelanos:
Reverted the revert. Quote pam_umask man page.

>        usergroups
>            If the user is not root and the username is the same as
>            primary group name, the umask group bits are set to be the
>            same as owner bits (examples: 022 -> 002, 077 -> 007).
> 
>        nousergroups
>            This is the direct opposite of the usergroups option
>            described above, which can be useful in case pam_umask has
>            been compiled with usergroups enabled by default and you want
>            to disable it at runtime.

Maybe that could fix the root issue.

adrelanos:
It might be possible to script Linux PAM so pam_umask does not apply to root.

adrelanos:
> It might be possible to script Linux PAM config so pam_umask does not apply to root.

Done.

adrelanos:
Readme was updated for pam_umask use.

monsieuremre:
> How would core dumps we accidentally enabled?

They wouldn't normally. But disabling them wherever possible is a net positive and brings no harm.

> Why is ulimit -c 0 extra, duplicated?

I don't understand how it's duplicated.

> Does not work. Files created by root still unreadable by others.

This is the expected funtionality. That's why we set it to 0027 or 0077. So that others can't read. I don't understand how this is a problem. Does it break anything? Root is arguably the most important user that should have this umask.

adrelanos:
> > How would core dumps we accidentally enabled?
> 
> They wouldn't normally. But disabling them wherever possible is a net positive and brings no harm.
> 
> > Why is ulimit -c 0 extra, duplicated?
> 
> I don't understand how it's duplicated.

Because core dump disabling is already disabled using the appropriate configuration files I see no benefit in doing it in /etc/profile.d again. Instead there's a bit harm:

* It gets more difficult for those who really need to enable core dumps to figure out how to re-enable it since now it's disabled in multiple places.
* Carrying unnecessary files.

> > Does not work. Files created by root still unreadable by others.
> 
> This is the expected funtionality. That's why we set it to 0027 or 0077. So that others can't read. I don't understand how this is a problem. Does it break anything? Root is arguably the most important user that should have this umask.

It breaks reading config files in /etc for:

*  tb-updater
* Firefox
* Thunderbird
* Probably other applications that run as user `user` but that read config files created by the system administator (or even packages) in `/etc`.

monsieuremre:
> It gets more difficult for those who really need to enable core dumps to figure out how to re-enable it since now it's disabled in multiple places.
It already is disabled in multiple places and they are documented. Anyone who needs coredumps would not go through this hardening. And if they choose to disable, we document all the places we disable coredumps.

> It breaks reading config files in /etc for

Permissions don't apply retroactively. So older files' permissions will stay the same. We can make exceptions for these files also for the future if they are countable, for example by manually granting read permission for others with the permission-hardener daemon. It is also an admin's responsibility to set the read permissions manually after creating configuration files. So manual stuff, no problem. Only the packages might be problematic. If these are too many to create exceptions for, we just stay with the current workaround.

adrelanos:
> > It gets more difficult for those who really need to enable core dumps to figure out how to re-enable it since now it's disabled in multiple places.
> > It already is disabled in multiple places and they are documented. Anyone who needs coredumps would not go through this hardening. And if they choose to disable, we document all the places we disable coredumps.
> 
> > It breaks reading config files in /etc for
> 
> Permissions don't apply retroactively.

Retroactive is relative. Once the new umask for root enabled, and after another package gets installed, the umask hardening might apply.

> We can make exceptions for these files also for the future if they are countable, for example by manually granting read permission for others with the permission-hardener daemon.

> It is also an admin's responsibility to set the read permissions manually after creating configuration files.

An admin trying to harden / preconfigure for example /etc/firefox-esr or /etc/thunderbird will have no clue whatsoever that the config files will be readable by root only but not by the actual applications to be configured. Since these are graphical applications, there might be no error message. If lucky, there might be a warning or error message from starting the application from a terminal emulator.

> So manual stuff, no problem. Only the packages might be problematic. If these are too many to create exceptions for, we just stay with the current workaround.

It cannot be guestimated how many such cases exist.

monsieuremre:
Hmm. Thinking. Yeah maybe. Current umask for everyone but root is already not that bad anyway. We can still add the coredump limit tho. I see no harm there. The use case of coredumps is really to my knowledge wastly limited to kernel development and debugging and stuff. Those people will figure how to disable coredumps if they choose to harden anyway, considering that would make their work harder.

adrelanos:
* https://github.com/Kicksecure/security-misc/issues/174

> How to even re-enable coredumps as of now?
> Is this implemented in debug-misc?
> 
> I don't want to configure us into a corner and then when somebody asks how to re-enable functionality, nobody knows the answer and it's a major effort to re-enable it.

monsieuremre:
I don't think we should worry about that much. Coredumps are of no use on a daily driver. Debugging and development are the only cases where someone would want to enable it, which I say, these people probably won't do kernel development or debugging on their daily driver. It is true that it might be useful for debugging and development in a general sense that is not limited to the kernel, but I don't think it is far fetched to assume these people are in a position to re-enable coredumps themselves.

adrelanos:
> but I don't think it is far fetched to assume these people are in a position to re-enable coredumps themselves.

Unfortunately that is a wrong assumption. I know, discussed similar things with skilled developers for C and whatnot but they would fail with the sysadmin part because that's a much different skill. Documentation exists on the internet but scattered.

Hence, my development standard has always been that that is hardened should be documented with known ways on how to undo it. If even you cannot easily answer above questions in much more time than required to type it that means that it's already too difficult.

The core dump might have been a slip through. Something I'll look into.

Complete, clean, documented, reversible solutions are welcome.

Otherwise incomplete, quick solutions with hardening only but without documentation and not easily reversible will take some time until I can complete them.

monsieuremre:
> Complete, clean, documented, reversible solutions are welcome.

Got it. Should not be difficult to document.

monsieuremre:
* Umask is already handled with pam for non-root users.

* Coredumps are already disabled and disabling within /etc is better left to the user.

Ergo, closing.

Issue #157, opened 2023-11-05
why remount-secure even needed? fix insecure mount options at the root, not here


adrelanos:
Which software / source code sets the initial insecure mount options anyhow?

Could we fix it there instead of adding a band-aid on top?

monsieuremre:
Well, actually there is a better way. A solution that is no band-aid. And it is modifiying the fstab. There is also another solution, that is direct and involves no remounting. We can write custom systemd mount targets and ship them with the package. The only problem with the latter is, fstab entries take precedence over those units. So we can't harden what is covered in fstab with this way.

monsieuremre:
Like if the user has a seperate /home volume, we can't harden it with our custom systemd unit. I think the reason behind this is, systemd dynamically overrides our unit with the unit it auto generates from fstab on startup.

adrelanos:
That doesn't really answer why it's insecure by Debian (or most Linux distributions) default.

monsieuremre:
Because by default debian and most distributions don't do too much disk partitioning. If you want a seperate tmp or var partition, most of the time you manually configure them when installing. And bind mounting to itself is clever workaround to get all the good mounting option benefits without having the partition itself, but it is really not that surprising that no distro would mount a directory to itself by default.

monsieuremre:
A very highly unrelated sidenote: the only distribution that I see who ships secure defaults by default is openSUSE. Debian by default ships with very slack file permissions for anything. OpenSUSE however ships with already mostly hardened defaults. Tumbleweed gets a score of 87 on lynis after a default install. Debian and most other distros get like 60. And this isn't surprising, because Debian does not try to be better desktop or server os. Debian tries to be the universal os, compatible with every device ever. So actually makes sense for them to ship slack defaults. A price to pay for universal compatibility.

adrelanos:
Where is the source code where it says "use these default mount options for this and that partition"?

adrelanos:
> Like if the user has a seperate /home volume, we can't harden it with our custom systemd unit. I think the reason behind this is, systemd dynamically overrides our unit with the unit it auto generates from fstab on startup.

Created https://github.com/Kicksecure/security-misc/issues/161 for it.

adrelanos:
quote https://lists.freedesktop.org/archives/systemd-devel/2019-December/043845.html

[...]

PR #165, opened 2023-11-17
Secure Remount Without Remounting At All For The Most Part - Secure Mount

monsieuremre:
This here. We don't own fstab, we don't modify fstab, we don't modify any files at all. We just add our own little config under respective directories. For var for examle under ```var.mount.d```. This targets the real partitions on disk. If there is literal partition for /home, when mounting it systemd gets everything from fstab, our options come on top of everrything afterwards. If there are no entries in fstab, so no partitions on disk, these overrides simply don't serve any purpose and are never read, so they do no harm. This makes sure we directly mount all partitions in fstab with the secure options in the first place. So no remounting occurs at all.

For directories that don't have their own partitions, we bind them. Still, no remounting. We bind these and we are done.

* No remounting
* Everything mounted with hardened options the first time
* Directories are bounded when necessary

So overall very good. The only caveat is, we can't apply our extra options for ```/run```, ```/dev``` and ```/dev/shm```. These can't be manipulated with mount unit overrides, they need to be addressed in fstab. 

``` Cannot create mount unit for API file system```

For these we have to find another solution, which is unfortunately going to be remounting probably. But for now, let's get everything else on board first. This solution covers everything and jumps no hoops. Everything is hardened from the beginning.

Please test. Also the old implementation has to be deleted before merging this.

### CODE REVIEW

**Comment by monsieuremre** [comment]
Actually I also took care of the ones that can't be overriden as well, just now. Untested, but should work. We only and only remount 2 things, ```/dev``` and ```/dev/shm```. These can't be overriden with systemd units. I think this is a massive improvement from remounting everything.

File: usr/bin/bind-directories [hunk 1]
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+if ! findmnt "/tmp"
+then
+  mount -o defaults,nodev,noexec,nosuid -t tmpfs tmpfs /tmp
+fi
+
+if ! findmnt "/var/tmp"
+then
+  mount -o defaults,nodev,noexec,nosuid --bind /tmp /var/tmp
```
adrelanos, line 10: > /tmp /var/tmp

This is a a bug? /tmp and /var/tmp have different purposes as per FHS.

usr/bin/bind-directories
```diff
@@ -0,0 +1,44 @@
...
+fi
+
+if ! findmnt "/var/log"
+then
+  mount -o defaults,nodev,noexec,nosuid --bind /var/log /var/log
+fi
+
+if ! findmnt "/boot/efi"
+then
+  if [ -d /boot/efi ]
```
adrelanos, line 25: So /boot/efi isn't a mountpoint but then check if /boot/efi is even an existing directly. But if this is a thing then this  test `-d` (is directory) should consistently used everywhere....

adrelanos: Better to use function `remount_secure` (modified) from `/usr/bin/remount-secure`?

* It checks if it's a mount point.
* It checks if the folder exists.
* Has debug output.

adrelanos:
Not sure yet. Many points to consider...

etc/systemd/system/boot-efi.mount.d/50_security-misc.conf
```diff
@@ -0,0 +1,2 @@
+[Mount]
+Options=nodev,noexec,nosuid
```
adrelanos, line 2: `Options=` would overwrite user /etc/fstab supplied options such as `noatime`? Quoted from systemd manual:

> Options=
> Mount options to use when mounting. This takes a comma-separated list of options. This setting is optional. Note that the usual specifier expansion is applied to this setting, literal percent characters should hence be written as "%%".

(`noatime` might not matter specifically for this mount point or not be the most important mount option but generally. Overwriting all user supplied options in /etc/fstab would be surprising and a bug.)

> Note that the usual specifier expansion is applied to this setting,

Maybe that helps to declare "add only options, don't remove options" (such as noatime)?

adrelanos:
As per this PR, the short summary:

* systemd mount files: For mount points declared in /etc/fstab.
* mount bind: For folders that do not have their own dedicated already existing mount point.
* remount: For /dev/ and /dev/shm.

Please correct and write a better one if I am wrong.

adrelanos:
At this stage with so many discussions scattered around various issues and PRs it would also be a highly useful contribution to have developer documentation summarizing all the various options and approaches as it gets increasingly convoluted and hard to reason about it. Could be added here:
https://www.kicksecure.com/wiki/Dev/remount-secure

adrelanos:
How is this PR better than...

* A) https://github.com/Kicksecure/security-misc/issues/157 (fix fstab in grml-debootstrap)
* B) The existing /usr/bin/remount-secure implementation?

This PR has systemd mount files, which is a nice approach in theory. But it gets harder to understand because in addition to the systemd mount files, two other mechanisms have to be used:

* mount bind: For folders that do not have their own dedicated already existing mount point.
* remount: For /dev/ and /dev/shm.

This makes it a lot harder to understand, predict what's happening, explain to users how to customize.

In practice, the systemd mount files would do little because the default fstab generated by grml-debootstrap is rather minimal. Therefore "lots of other stuff" is going on in /usr/bin/bind-directories, namely

* /tmp
* /var/tmp
* /var
* /var/log
* /boot/efi
* /boot
* /usr/share

In addition to that there's also /usr/bin/remount-api with an almost duplicated systemd unit.

Therefore I am asking, how is this PR better than the existing /usr/bin/remount-secure implementation? It can handle all of the above mentioned mount points and in addition to that has the following advantages:

* 1 systemd service unit file
* in 1 script,
* with 1 mechanism (mount) (shell function chooses or omits remount option depending on folder already existing or not),
* is configurable [1]
* has debug output [1]
* protected against double mounts (when restarting the systemd unit / re-running the script) [1]

----

Why not use systemd mount files for everything including bind mounts?

This article says it's possible:
https://utcc.utoronto.ca/~cks/space/blog/linux/SystemdBindMountUnits

Could it be an issue that we don't know if it's a dedicated mount point or just a folder? Could this be phrased a missing systemd feature / feature request? In that case, please point to or draft/write a systemd feature request.

Let's take for example /var.

* X) I guess in most cases it's just a folder. -> bind mount required.
* Y) Or it's on some systems with custom /etc/fstab a dedicated mount moint. -> remount required

The existing /usr/bin/remount-secure implementation would handle both cases, X and Y.

This PR would as of now only handle case X.

----

[1] These features could be dropped / replaced by consistently using systemd mount files for everything. [2]
[2] Maybe with an exception for /dev and /dev/shm, not ideal, but ok.

usr/bin/bind-directories
monsieuremre: This is not a bug. It is intentional.

Normally on servers and professional deployments it is not uncommon to have ```/var/tmp``` as a seperate partition. This aims to prevent ```/var/tmp``` being filled arbitrarily and stop ```/var``` from functioning by physically throttling space. On desktop computers, a seperate partition is for the most part unnecessary and costs valuable diskspace.

Binding ```/var/tmp``` to ```/tmp``` addresses this issue on systems that do not want to dedicate the disk space but still want to have the extra benefit of not risking to throttle ```/var```. It is not uncommoon practice to bind these two and mount them on ram in high security environments.

It is recommended to do this by CISOfy, if ```/var/tmp``` is not on a seperate partition. See:
# CISOfy - Lynis
# https://github.com/CISOfy/lynis/blob/master/include/tests_filesystems
# Test        : FILE-6376
# Description : Bind mount the /var/tmp directory to /tmp

It also makes sure the secure mount options are applied to this section as well without having to bind to itself. Temp files residing in tmpfs is also a security benefit for the desktop user. It should bring up no complications or breakages.

usr/bin/bind-directories
monsieuremre: > So /boot/efi isn't a mountpoint but then check if /boot/efi is even an existing directly. But if this is a thing then this test -d (is directory) should consistently used everywhere....

I had the same thought but then concluded, no. This is the only thing that might not exist at all. For everything else, checking directories is comletely superfluous.

For examle, if there is no partition for ```/boot```, then ```/boot``` is a directory on the root pratition. There is no other possibility. So it is totally unnecessary to check if ```/boot``` is a directory. Because it is. It has to be. If it is no partition, then it is a directory. No system can exist without it.

```/boot/efi``` is the only exception here. If there is no partition for it, there is a good chance it might not exist at all. Because it is not universal in the same sense. It exists only in UEFI system. If a system runs on legacy BIOS, there won't be ```/boot/efi```. So here is the only place it makes sense we check if this directory exists at all.

Normally ```/boot/efi``` is almost always a seperate partition on UEFI systems because it has the ```vfat``` file system. Since most linux installs default to ```ext4``` or something else for the root partition, they do a seperate partition for this. But it seems also possible to have everything in ```vfat``` and not use anything else, thus not having a seperate partition for ```/boot/efi``` but rather having it as a directory. This is why we check, for this possibility.

File: etc/systemd/system/boot-efi.mount.d/50_security-misc.conf (hunk 3 above)
Comment (monsieuremre) [reply etc/systemd/system/boot-efi.mount.d/50_security-misc.conf:2]: > Options= would overwrite user /etc/fstab supplied options such as noatime?

Absolutely no. I tested several combinations. In any case, these options are only added to the existing one. ```noatime``` or anything else will not be overridden, unless we provide an option that would contradict it. Like if fstab has ```suid``` and if we give ```nosuid```, then ours will take precedence. But if the options are not mutually exclusive, fstab options just stay there as they are.

So this is not a problem.

**Comment by monsieuremre** [comment]
> As per this PR, the short summary:
>    * systemd mount files: For mount points declared in /etc/fstab.
>    * mount bind: For folders that do not have their own dedicated already existing mount point.
>    * remount: For /dev/ and /dev/shm.
>
> Please correct and write a better one if I am wrong.

A yessir you summarized very well.

> How is this PR better than... B) The existing /usr/bin/remount-secure implementation?

First and foremost: it jumps no hoops. The existing solution remounts everything. This on its own:
* Consumes considerable time on boot
* Adds unnecessary complication and risks breakage, because a file might be in use on the partition while we remount
* Leaves a time frame where partitions are mounted with insecure options
* Is not direct and only applied after the fact

This solution here is:
* Direct
* Makes sure partitions are mounted with secured options in the first place and not remounted
* Saves time, since no remounting
* Reduces breakage and complexity, since there is no chance to of mounting while using the file in the partition
* Does not require dynamically checking or adjusting stuff
* Literally does not execute anything at all (the config files that make sure everything mounts securely in the first place)

They only resemble each other when it comes to remounting ```/dev``` and ```/dev/shm```. But since these are not real partitions and rather quasi partitions that actually reside on ram, it is infinetely faster to remount them. It still is not as good as our solution for the other partitions, but these can't have that for reasons I explained. But this is a rather small compromise.

For binding directories, they are the same in outcome but definetely not the same in functionality. The older implementation is designed to remount things and maybe bind them on demand, has more complexity. This one is much simpler and just bind the directories.

The remount-api is also very, very simple.

And I see there is some misunderstanding here so let me clear it.

> Could it be an issue that we don't know if it's a dedicated mount point or just a folder? Could this be phrased a missing systemd > feature / feature request? In that case, please point to or draft/write a systemd feature request.
>
> Let's take for example /var.
>
>    X) I guess in most cases it's just a folder. -> bind mount required.
>    Y) Or it's on some systems with custom /etc/fstab a dedicated mount moint. -> remount required
>
> The existing /usr/bin/remount-secure implementation would handle both cases, X and Y.
>
> This PR would as of now only handle case X.

No. This is not a problem. You explained correctly what the old implementation does. So let me explain what my new implementation does.

 Let's take for example /var.

    X) Just a directory -> Bind
    Y) Is partition -> Do nothing

I think you can see the difference. We don't take the effort to remount it or whatever. Because we know that it mounts with the correct options anyway, in the first place. What we check is if it is a directory, we bind. That's it. We do nothing because when systemd generates the mount units using fstab, our config files come on top of those units and make sure the units have hardened options in the very start.

On our end, we can literally also do nothing and everything will be mounted securely. Binding stuff is just something extra we do for directories.

So in fact, I'd say the implementations are vastly different from the ground up.

**Comment by monsieuremre** [comment]
And also, I can split this pull into two so that the config files are merged first, because they are literally just files and don't execute anything and cannot possibly break anything.

The remounting of only /dev and /dev/shm and binding of non partitions is also still considerably faster/simpler in comparison.

File: usr/bin/bind-directories (hunk 2 above)
Comment (adrelanos) [reply usr/bin/bind-directories:25]: It's a very valid use case to have a system without `/boot`.
(KVM Linux host operating system provided direct kernel boot, a Linux distribution such as [Kicksecure inside a folder (chroot) using `systemd-nspawn`](https://www.kicksecure.com/wiki/Chroot), OpenVZ probably and others)

File: usr/bin/bind-directories (hunk 1 above)
Comment (adrelanos) [reply usr/bin/bind-directories:10]: This could use a comment with a few links to references (most not being big source code files).

**Comment by adrelanos** [comment]
note to self:

* check if noatime etc. in fstab is preserved during review

**Comment by adrelanos** [comment]
To re-iterate my question from above (see with more content surrounding it):

> Why not use systemd mount files for everything including bind mounts?

**Comment by adrelanos** [comment]
> > How is this PR better than... B) The existing /usr/bin/remount-secure implementation?
> 
> First and foremost: it jumps no hoops. The existing solution remounts everything. This on its own:
> 
> * Consumes considerable time on boot

Why would this PR be faster? Since a default /etc/fstab (auto generated during build process by grml-debootstrap) is non-hardened), the systemd mount drop-in config snippets do effectively nothing.

> * Adds unnecessary complication and risks breakage, because a file might be in use on the partition while we remount

Same issue with /usr/bin/bind-directories as per this PR.

This can - maybe - only be fixed by the correct systemd unit files ordering, i.e. perhaps by using [sysinit-post.target](https://github.com/Kicksecure/security-misc/blob/master/lib/systemd/system/sysinit-post.target). (Not yet tested.)

> * Leaves a time frame where partitions are mounted with insecure options

Effectivly the same as per this PR. Same reasoning as above in this very comment.

> * Is not direct and only applied after the fact
> 
> This solution here is:
> 
> * Direct
> * Makes sure partitions are mounted with secured options in the first place and not remounted
> * Saves time, since no remounting
> * Reduces breakage and complexity, since there is no chance to of mounting while using the file in the partition

In theory. If there were partitions, but there aren't by default in a Kicksecure VM.

> * Does not require dynamically checking or adjusting stuff

I don't see this as an issue.

This PR too needs to check:

* is mountpoint
* folder exists (/boot and /boot/efi)

This PR combines as of now the complexity of systemd drop-in files with the systemd service unit file + shell scripting approach. Best to settle for 1 approach only.

That's why I suggested to use use systemd for bind mounts too if that is possible. If it's all systemd [1] drop-in files that would be nice. Then it's worth to avoid the complexity of shell scripting.

[1] With the unfortunate exception of /dev and /dev/shm.

> * Literally does not execute anything at all (the config files that make sure everything mounts securely in the first place)

If it was systemd mount drop-in files only, yes.

> They only resemble each other when it comes to remounting `/dev` and `/dev/shm`.

And all or mostly the same bind mounts as per /usr/bin/bind-directories.

> For binding directories, they are the same in outcome but definetely not the same in functionality. The older implementation is designed to remount things and maybe bind them on demand, has more complexity. This one is much simpler and just bind the directories.

The currently existing implementation has less complexity because it doesn't combine systemd mount drop-in files the shell scripting approach. Try to explain to users how to customize that. Gets totally messy to explain. The current implementation also has nice journal output which a system administrator can review to see what's going on.

> I think you can see the difference. We don't take the effort to remount it or whatever. Because we know that it mounts with the correct options anyway, in the first place. What we check is if it is a directory, we bind. That's it. We do nothing because when systemd generates the mount units using fstab, our config files come on top of those units and make sure the units have hardened options in the very start.

I see.

Major next point: Please try using systemd mount drop-in files for the bind mounts too.

* If unsure how to do it, please ask  on systemd-devel mailing list (or prepare a draft here for me to post there).
* If there's a missing systemd feature for bind mounts, please point to an existing feature request, draft one here or post one. This is to ensure that this is really impossible using systemd mount drop-in files.

I'd prefer a mostly [1] systemd mount drop-in files based implementation without any shell scripting as this has a good chance of being (relatively) bug free, easy to understand for system administrators, advanced users, auditors (because systemd documentation will be applicable, systemd upstream bug reporting will be possible, no shell scripting) and easily customizeable with lexically higher systemd mount drop-in files.

**Comment by monsieuremre** [comment]
> Major next point: Please try using systemd mount drop-in files for the bind mounts too.

Will try. This was my initial intention anyway. The problem that occured was, where to create these mount units. If I create them in /etc, they would override the fstab and the partitions would not be mounted. I am guessing that if I create them under /usr/lib, it might work fine. Will try and report.

> I'd prefer a mostly [1] systemd mount drop-in files based implementation without any shell scripting

This won't be possible even if I get the first point right now, because of /dev/shm and /dev. For these, the only way is to edit fstab or remount. Actually it might be possible, but I don't know how. Because on a normal system these are mounted with the default options. Debian partially hardens them as default. So there is a way to edit the default options for these file systems. It seems to be a kernel parameter or something similar that is set in compile time. I don't know what this is, but if I find out, we might be able to tackle this. There is unfortunately not that many resources so I will see what can be done if anything.

**Comment by monsieuremre** [comment]
And I did exactly that. I tested the following scenario:

There is a /home partition, there is no /var partition. Mount units for home and var are created as they are here under /usr/lib. What happens:
* var gets bound to itself securely
* home unit does not activate, because it gets overriden by fstab

So just as we want. The others I haven't tested just yet, but I see no reason we can't generalize this result. No script no exection no nothing.

**Comment by monsieuremre** [comment]
Done some big research on how we can handle /dev and /dev/shm. And turns out, there was a fstab.d in the past. It was added to util-linux, but was reverted afterwards. See [here](https://marc.info/?l=util-linux-ng&m=132740311801201&w=2). The --fstab options supports defining more than one fstab files.

Fortunately, I found a way to not remount anything. Everything will be mounted securely. And I also a found a way to define a file to add to the existing fstab dynamically.

Unfortunately, I actually didn't because it requires systemd version 254, and debian bookworm ships with 252. But apart from that, the solution is there.

What we do is, we set the following kernel parameter:
systemd.mount-extra=tmpfs:/dev/shm[:tmpfs[:defaults,noexec,nosuid,nodev]]

And that's kinda it. Now, since we can't do this because of the version, I am looking for another solution.

**Comment by adrelanos** [comment]
> > Major next point: Please try using systemd mount drop-in files for the bind mounts too.
> 
> Will try. This was my initial intention anyway. The problem that occured was, where to create these mount units. If I create them in /etc, they would override the fstab and the partitions would not be mounted. I am guessing that if I create them under /usr/lib, it might work fine. Will try and report.

Or ConditionPathExists ConditionPathIsDirectory or similar as per https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html#ConditionPathExists= as mentioned in a prior ticket or PR.

> > I'd prefer a mostly [1] systemd mount drop-in files based implementation without any shell scripting
> 
> This won't be possible even if I get the first point right now, because of /dev/shm and /dev.

As mentioned, these 2 can remain an unfortunate exception.

> For these, the only way is to edit fstab or remount. Actually it might be possible, but I don't know how. Because on a normal system these are mounted with the default options. Debian partially hardens them as default. So there is a way to edit the default options for these file systems. It seems to be a kernel parameter or something similar that is set in compile time. I don't know what this is, but if I find out, we might be able to tackle this. There is unfortunately not that many resources so I will see what can be done if anything.

Please draft for or ask systemd-devel mailing list.
(As the refusal message to remount seems to be hardcoded in systemd.)

**Comment by adrelanos** [comment]
Wrote my comment before reading your follow-up comments and new git changes.

**Comment by adrelanos** [comment]
So this is starting to look very nice.

----

* /lib/systemd/system/remount-api.service is unfortunate but acceptable. Has a good chance of not breaking anything, I guess.
  * Also if in Debian trixie and above this can be replaced by kernel command line `systemd.mount-extra` then this implementation would become even cleaner.
  * Could you please add a /etc/default/grub.d drop-in files but commented-out by default with the `systemd.mount-extra` options and a comment that this should be used in Debian trixie and above?

----

Could you please rename all `50_` to `30_`? This is because in Kicksecure we usually documented `50_user.conf` to be used by users.

**Comment by adrelanos** [comment]
Other todo: testing.

Please let me know once this is tested.

Readme:

* Could you please add this to readme?
* Please mention why sometimes /etc vs sometimes /lib is being used.

**Comment by monsieuremre** [comment]
> Could you please add this to readme?
> Please mention why sometimes /etc vs sometimes /lib is being used.

Yeah sure. Better done after merging tho, since things might change on the go. It is because what takes precedence over what.

**Comment by adrelanos** [comment]
I'd also like to see:

```
[Unit]
ConditionKernelCommandLine=!noremount
```

but I could do that after merge. (The exact name of the kernel parameter can be discussed.)

This is to make it simple to test this with/without and to make it easy to disable all of this in case there are unpredictable issues.

monsieuremre:
The thing is, I haven't even tested the command line parameter yet on debian sid. The documentation makes it seem like it will work for sure.

And I am not really motivated to test just yet. First I want to tackle it in the current version. Then for the next releases it might just become even cleaner.

When it comes to remount. /proc is also an api file system. I want to remount this one too, but with hidepid=4 and all the good old options. I think we can merge the existing hidepid service inside this one to reduce the number of services. Should I proceed?

adrelanos:
> When it comes to remount. /proc is also an api file system. I want to remount this one too, but with hidepid=4 and all the good old options. I think we can merge the existing hidepid service inside this one to reduce the number of services. Should I proceed?

This can be done later after this one was tested and merged. It already 
has its own dedicated ticket.

Not mixing too much stuff together which has potential for breakage.

monsieuremre:
Ok. The drop in configs under /etc work perfectly in any case all the time and they make sure every partition is moutned securely.

The same can't be said about bind units. There are some certain problems. Double mounts are unavoidable with /var/log.

I think we shouldmerge the first part now, since it works. The bind and /dev /dev/shm, I will handle seperately. Does this sound ok?

adrelanos:
No, I don't want to mix two different approaches, that is systemd drop-ins for partitions mixed with shell scripting for bind mounts. Too complex to understand, document, customize.

The current implementation (remount-secure systemd unit and script) is less complex and covers all. But even that is non-ideal.

After all the time spent on this topic, all issues and corner cases with all the hacks that aren't directly shipping a hardened /etc/fstab by default I think that https://github.com/Kicksecure/security-misc/issues/157 (i.e. fixing this issue at the root in image creation tools / installers such as grml-debootstrap / calamares) seems the most promising, stable, simple.

monsieuremre:
Well this is the day. At last. Today we harden them all, no remounting at all, not even ```/dev/shm``` or anything, and also no changing or modifying or owning the /etc/fstab.

I found a way to implement a quasi fstab.d manually. Very easy, works, and it is epic. I tested it too. Here are the steps:

* First we have to redefine the fstab path. For that we set the following kernel parameters:
* GRUB_CMDLINE_LINUX_DEFAULT="LIBMOUNT_FSTAB=/etc/something SYSTEMD_SYSROOT_FSTAB=/etc/something SYSTEMD_FSTAB=/etc/something"
* Then, magic is done. Now ```/etc/something``` is our new fstab file, and we ignore ```/etc/fstab``` completely. Now, the only problem here is, if there is a change in ```/etc/fstab```, we want to respect these changes. One way to do this is always dynamically generate the hardened fstab from the old fstab with a very very early hook. Systemd generators run very early, right after the init process. So we can't do this generation with a service, it has to be a hook. So basically we do this:
* We dynamically generate the fstab ```/etc/something``` using the original ```/etc/fstab``` with a hook.
* We copy the existing fstab contents into /etc/something and add our extra lines for ```/dev/shm``` and stuff.

We do not touch or own the old ```/etc/fstab```. The user can still modify this. Please consider also the following points:

* We never call mount, ever
* Systemd handels everythin as usual
* We never remount
* Everything mounts securely in the first place
* We never manually check stuff
* All we do is generate a custom fstab from the original one, without modifying or owning it.

Would this be a desirable option? I did a lot of research and I can tell with confidence that it won't get any better than this probably.

adrelanos:
That is for sure an interesting, clever approach but maybe not the robust, stable and perhaps boring (?) implementation I am looking for.

> * GRUB_CMDLINE_LINUX_DEFAULT="LIBMOUNT_FSTAB=/etc/something SYSTEMD_SYSROOT_FSTAB=/etc/something SYSTEMD_FSTAB=/etc/something"

The mostly minimal kernel command line modification is nice to avoid bloating it even more as the other security-misc kernel parameters are already expanding it a lot.

> * Then, magic is done. Now `/etc/something` is our new fstab file, and we ignore `/etc/fstab` completely. Now, the only problem here is, if there is a change in `/etc/fstab`, we want to respect these changes.

> Systemd generators run very early, right after the init process.

But aren't these controlled by systemd unit files too?

> So we can't do this generation with a service, it has to be a hook.

Hook? Dracut hook?

(Implementation detail: The "generate our fstab" script should be separate so it can be easily run/debugged.

> * We dynamically generate the fstab `/etc/something` using the original `/etc/fstab` with a hook.

`awk`, `sed`, `str_replace`, `bash` `while true` `read` loop? That would be the difficult, fragile, messy part.

> * We copy the existing fstab contents into /etc/something and add our extra lines for `/dev/shm` and stuff.

For `/dev/shm` and `/dev` this seems good. These are just simple appends. But for real partitions and remount not so great. Issues are:

* messy custom fstab auto generation
* if the user edits /etc/fstab and uses mount, it would reset the hardened mount options? Also be confusing for users why mount options are hardened in the first place because these aren't done in the place where everyone would except these. That is in /etc/fstab.
* configurability: how to disable /tmp noexec or how to disable/enable /home noexec. That would still be confusing for users since based on the custom fstab auto generator.

> We do not touch or own the old `/etc/fstab`. The user can still modify this. Please consider also the following points:
> 
> * We never call mount, ever
> * Systemd handels everythin as usual
> * We never remount
> * Everything mounts securely in the first place
> * We never manually check stuff
> * All we do is generate a custom fstab from the original one, without modifying or owning it.

This is good.

> Would this be a desirable option? I did a lot of research and I can tell with confidence that it won't get any better than this probably.

It seems to me you're into an intellectual challenge, i.e. like to solve complex issues by coming up with more complex solutions but I am confused why https://github.com/Kicksecure/security-misc/issues/157#issuecomment-1817343098 is continuously ignored.

I don't think a better solution can be found than shipping a hardened by default /etc/fstab in newly built Kicksecure images. That seems the most simple and robust solution.

monsieuremre:
> I don't think a better solution can be found than shipping a hardened by default /etc/fstab in newly built Kicksecure images. That seems the most simple and robust solution.

I skip over this one because I would rather have this package universally harden the mounting options even on a debian install. And I think this is possible too.

> Hook? Dracut hook?

Init hook. Not depending on dracut, more preferably.

> awk, sed, str_replace, bash while true read loop? That would be the difficult, fragile, messy part.

Now, I hear your point about messy fstab generations. But I have to disagree. This is no more messy than having mount units, in fact less messy I would say. We can do this:
* Leave the drop in configs for already existing partitions under /etc, these work perfectly always.
* Use the new fstab generation but do not modify lines or anything with complex operations. We will just at boot, merge the two fstab files (```/etc/fstab``` and ```/etc/our_thing``` to ```/etc/new_fstab```) together and use it as the fstab file. I'm talking about an init hook which just runs ```cat /etc/fstab /etc/our_thing > /etc/new_fstab```. That's it.

By this we achieve:
* Hardened configs for already existing partitions with just drop in config files.
* No sed, awk inline modification or anything to the fstab.
* Hardened moutns for api file systems like ```/run```, ```/dev``` and ```/dev/shm```.

Now, we just have to handle binds. For this we can parse the old fstab and add the necessary bind lines dynamically. This won't/can't be messy. Because fstab already has a predefined format. Parsing can't go wrong. And after that, we just add extra lines, which also can't go wrong.

I think this one you would be more accepting. I eliminated the modification of lines completely. There is just adding new lines left. What would you say to this option?

monsieuremre:
Alright. Then closing this one. I suppose the old remount secure service should be removed from the package and the readme as well in this case.

monsieuremre:
I want to re open this pull. There was only one little thing that was preventing this branch from working completely and only with drop in config files, and that was the api file systems like ```/dev``` and ```/proc```. From systemd version 254 and onwards, we can harden these with mere drop in config files too, by setting kernel parameters. Debian 12 uses version 252, but I did some testing on sid. It works obviously.

I have to say, this option is too good to discard. It is compatible with all fstab files and all partitioning schemes. There is also no work done on our end, just some lines in some files. The newer systemd versions allow filling the last missing part. Normally, we can already achieve these:

* Making sure already existing mounting takes place with hardened options.
* When no unit generated by fstab and/or no user unit, our custom bind unit does binding. If anything else is already there, it takes precedence.

Our missing piece was:
* Mounting API filesystems with hardened options. Namely we want to modify ```/dev``` and ```/dev/shm```

Our missing piece is possible with a kernel parameter. I say this. Speed is of essence. Let's do the hardening. Any time there is a choice between uncertainty in the future vs. certainty now, we gotta choose now. I want to re-open this request if you are willing to take this path @adrelanos.

adrelanos:
The branch was deleted by you so I cannot use the reopen button.

monsieuremre:
I will restore the branch. Are you open to this idea of using config files only and being fully effective after the next debian release? If you say it might be 'considered', that would be enough for me to restore the branch. But I do not want to create a distraction if you are certain this is not way you want to address this issue.

adrelanos:
Yes.

monsieuremre:
I've done some contributions to the wiki. To some various pages including this one. If there was a way to make wiki contributions doing a pull request here, it would be easier to manage and easier to track. I don't know how difficult this would be to implement, it's just some unrelated feedback that you might consider.

monsieuremre:
> Yes.

Ok good. Soon when I find the time I will restore the branch, check everything again and reopen the pull.

adrelanos:
> I've done some contributions to the wiki. To some various pages including this one. If there was a way to make wiki contributions doing a pull request here, it would be easier to manage and easier to track. I don't know how difficult this would be to implement, it's just some unrelated feedback that you might consider.

MediaWiki content fortunately doesn't have first class support for git integration otherwise I would be all over it. As a user of both, MediaWiki and git this is on top to my tooling wishlist. To have git based updates and also have a bridge for those who prefer to edit locally, use git branches, pull requests, etc. There's git-mediawiki but not well maintained last time I checked so that unfortunately broke.

related:
https://www.kicksecure.com/wiki/Dev/About_Infrastructure#MediaWiki

adrelanos:
> > Yes.
> 
> Ok good. Soon when I find the time I will restore the branch, check everything again and reopen the pull.

That was done with https://github.com/Kicksecure/security-misc/pull/195? A bit confusing to have a new pull request.

monsieuremre:
Yes absolutely. The new request is literally the same content so practically the same as restoring the branch. I did it for it to be more clean.

Issue #154, opened 2023-11-04
Scope of application-specific hardening?


adrelanos:
https://github.com/Kicksecure/security-misc/pull/146 (which looks nice at first sight) and [1500 AppArmor profiles](https://apparmor.pujol.io/) (impressive scope) woke me up to set the expectations straight for [application-specific hardening](https://github.com/Kicksecure/security-misc#application-specific-hardening).

How many applications should be hardened through this repository?

Maybe best to limit security-misc to "global" / "system" wide hardening, have a separate repository for application specific hardening?

Or I could arbitrarily limit the applications to those pre-installed in Kicksecure, Whonix and perhaps some other popular/important applications (even more arbitrary).

However,  ~1500 pull requests for all sorts of applications hardening I've never used and reviewing the details of this with DOS the development so this isn't possible.

These expectations need to be set straight in the readme to respect contributor's time.

For a hypothetical ~1500 application hardening settings a separate repository would need to be maintained by somebody else.

monsieuremre:
How about limit the applicaiton specific hardening to those that are default installed kicksecure/whonix applications. Thunderbird hardening was already in the package, I just added some more lines. We can just limit this manual application specific hardening to:
* Debian Specific System Components (Pretty much just apt probably)
* System Services (Like enabling NetworkManager IPv6 privacy options and enabling MAC address randomization)
* Freedesktop (Like forcing Xorg to run as non-root, which I was planning to create a pull on)
* Kicksecure default user applications (Thunderbird, VLC, KeepassXC, etc.)

adrelanos:
Sounds good!

Can be added to readme.

> * Freedesktop (Like forcing Xorg to run as non-root, which I was planning to create a pull on)

Not sure that's worth it. Got any input on the following ones?

https://forums.whonix.org/t/port-to-wayland/17380

https://forums.whonix.org/t/custom-desktop-environment-wayland-support/17490

adrelanos:
Expanded.

PR #166, opened 2023-11-17
USB Guard | Depend on it and configure rules

monsieuremre:
We should depend on usbguard as a package. When installed on debian, usb guard settings are configured in a way that all devices are rejected, only those that were already plugged in before the daemon starts are allowed.

Normally this is not optimal. We should by default reject everything and only whitelist the devices that we know and trust. But this requires configuration at the user end. So for now, we better leave the debian config as is.

We also add new rules with our config file. These rules do not configure the daemon. They define how the daemon behaves when it is 'applying policy'. The policy is by default blocking everything. We slacken this by allowing devices with mass storage interface, so the users can plug in their external hard drives and usb's. Then we blacklist especially suspicuous storage devices. Those that try to also behave like a keyboard at the same time for example.

Normally this is also not optimal, but it is a big improvement. We don't allow anything else. That means everything else is implicitly blocked. If the user plugs in a keyboard or mouse or anything else, it will be blocked. If these devices are plugged in when the machine starts up, they are allowed. This should not be too inconvenient for the user, because most peripherals are almost always already plugged in when the machine starts up. If not, they can be manually allowed or the system can be rebooted for the device to be allowed after having booted.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -30,6 +30,7 @@ Depends: adduser,
          python3,
          secure-delete,
          sudo,
+         usbguard,
          ${misc:Depends}
 Replaces: anon-gpg-tweaks, swappiness-lowest, tcp-timestamps-disable
 Description: Enhances Miscellaneous Security Settings
--- a/etc/usbguard/rules.d/30_security-misc.conf
+++ b/etc/usbguard/rules.d/30_security-misc.conf
@@ -1,0 +1,18 @@
+## Blacklisting is not the optimal approach to security. Normally all USB devices should be rejected (default) and only the devices you personally know and trust be whitelisted.
+## If you can do this, it is recommended to do. For convenience for the majority, we do the following:
+
+## Allow all USB devices with mass storage interface
+allow with-interface equals { 08:*:* }
+
+## Reject storage devices that also have extra suspicuous interfaces.
+## Like a usb storage device that also tries to behave like a keyboard.
+## This is a well known type of cyber attack.
+
+reject with-interface all-of { 08:*:* 03:00:* }
+reject with-interface all-of { 08:*:* 03:01:* }
+reject with-interface all-of { 08:*:* e0:*:* }
+reject with-interface all-of { 08:*:* 0a:*:* }
+reject with-interface all-of { 08:*:* 02:*:* }
+
+## We do not allow anything else. Keyboards, mice, and everything else, they will be rejected. The only exception is, if they were plugged in when the daemon starts.
+## If you have your keyboard plugged in before booting, it will be allowed. If you plug after the fact, you have to manually allow the device or do a restart.


adrelanos:
https://forums.whonix.org/t/usbguard-on-kicksecure-to-prevent-hardware-keyloggers-badusb/11988

adrelanos:
* https://github.com/Kicksecure/security-misc/issues/169

adrelanos:
"`Depends: usbguard`" will be added in source code repository `kicksecure-meta-packages`. Maybe to `kicksecure-cli-host-packages-recommended`.

The config files are good to have here.

related:

* https://github.com/Kicksecure/security-misc#project-scope-of-application-specific-hardening
* https://github.com/Kicksecure/security-misc/issues/169
* https://github.com/Kicksecure/security-misc/issues/154

monsieuremre:
So @adrelanos what are waiting regarding this?

adrelanos:
1. please removed the `Depends:`
2. please add a link to the source: https://usbguard.github.io/documentation/rule-language
3. the following comment seems wrong.

```
## Allow all USB devices with mass storage interface
allow with-interface equals { 08:*:* }
```

This actually does something else as per documentation.

> This policy will block any device that isn’t just a mass storage device. Devices with a hidden keyboard interface in a USB flash disk will be blocked. Only devices with a single mass storage interface will be allowed to interact with the operating system.

Best to copy this verbatim?

4. Where did you find the following one?

```
reject with-interface all-of { 08:*:* 0a:*:* }
```

monsieuremre:
> Where did you find the following one?

The rules aren't necessarily taken from a source. Some of them I wrote myself. It means: reject a device, that presents itself as a mass stroage device, and a CDC data device, at the same time.

Class codes for different interfaces can be looked up easily [here](https://www.usb.org/defined-class-codes).

etc/usbguard/rules.d/30_security-misc.conf
```diff
@@ -0,0 +1,46 @@
+# We allow those that were plugged in before the daemon starts. Everything is blocked as the default. Following rules apply on top of this.
+
+# Explicitly reject any interface that is not documented and/or defined by USB.org
+# Note: Most probably superfluous
+reject with-interface none-of { 01:*:* 02:*:* 03:*:* 04:*:* 05:*:* 06:*:* 07:*:* 08:*:* 09:*:* 0a:*:* 0b:*:* 0d:*:* 0e:*:* 0f:*:* 10:*:* 11:*:* 12:*:* 13:*:* 14:*:* 3c:*:* dc:*:* e0:*:* ef:*:* fe:*:* ff:*:*}
```
ArrayBolt3, line 5: According to https://usb.org/defined-class-codes, there is no base class of `04`, so that should be removed from the list. There also is a base class `00` for devices where class information is supposed to be determined from their interface descriptors, so that maybe should be added here. (If we're blocking everything by default anyway and this is superfluous, it probably won't harm security at least.)

Also, you seem to be missing a space before the closing `}` here. Not sure if that will result in a syntax error or not.

etc/usbguard/rules.d/30_security-misc.conf
```diff
@@ -0,0 +1,46 @@
...
+
+### Allow all mouses and keyboards, in a sense, so the user can conveniently change them without restrating the daemon.
+### Take extra measures to ensure security
+
+# Allow only one keyboard to be connected
+allow with-interface one-of { 03:00:01 03:01:01 } if !allowed-matches(with-interface one-of { 03:00:01 03:01:01 })
+
+# Allow only one mouse to be connected
+allow with-interface one-of { 03:00:02 03:01:02 } if !allowed-matches(with-interface one-of { 03:00:02 03:01:02 })
```
ArrayBolt3, line 14: According to the USB HID spec (https://usb.org/sites/default/files/hid1_11.pdf), there's a couple of invalid rules here - `03:00:01` and `03:00:02` will never exist, the only `03:00:*` devices that should exist are `03:00:00` devices. Quoting the spec:

> A variety of protocols are supported HID devices. The bInterfaceProtocol member of an Interface descriptor only has meaning if the bInterfaceSubClass member declares that the device supports a boot interface, otherwise it is 0.

bInterfaceProtocol is this last byte of the triplet, while bInterfaceSubClass is the byte in the middle. When the byte in the middle is `00`, it is *not* declaring that the device supports a boot interface, thus the last byte will be `00`.

Worthy of note, this rule will prohibit mice and keyboards that don't have boot interface support (I think my Logitech MX devices fall into this category), and it will probably prohibit HID devices other than mice and keyboards such as touchscreens. I don't know how common mice and keyboards without boot interface support are though. We should probably document how to disable this rule for people with problematic devices.

etc/usbguard/rules.d/30_security-misc.conf
```diff
@@ -0,0 +1,46 @@
...
+
+# Explicitly reject any device with a mouse/keyboard interface in combination with some other interface
+# Mouses and keyboards should only have one interface for all legitimate use cases
```
ArrayBolt3, line 17: Technically, mouse and keyboard devices *might* legitimately have two or even three interfaces, in the case of a wireless mouse+keyboard combo with a single receiver. The Microsoft Sculpt desktop set probably falls into this category, it consists of a keyboard, mouse, and touchpad, all separate but all sharing one receiver. So this second line of the comment isn't really accurate from a factual standpoint.

The second line of the comment also fails to reflect what the rules do. It prohibits mixing HID *and other* interfaces, it notably does *not* prohibit mixing multiple HID interfaces together. This is actually good IMO, but it's not what the comment says, so I think this second line should just be removed, or perhaps replaced with:

`# Mice and keyboards should likely never have non-HID interfaces provided alongside them`

If this line is kept, `mouses` should change to `mice` here also.

etc/usbguard/rules.d/30_security-misc.conf
```diff
@@ -0,0 +1,46 @@
...
+reject with-interface all-of { 03:*:* 02:*:* }
+reject with-interface all-of { 03:*:* 04:*:* }
```
ArrayBolt3, line 19: Here again there is no base class of `04`, so this line should be removed. Also, this list misses the `00` and `01` base classes at the start.

etc/usbguard/rules.d/30_security-misc.conf
```diff
@@ -0,0 +1,46 @@
+# We allow those that were plugged in before the daemon starts. Everything is blocked as the default. Following rules apply on top of this.
+
+# Explicitly reject any interface that is not documented and/or defined by USB.org
+# Note: Most probably superfluous
+reject with-interface none-of { 01:*:* 02:*:* 03:*:* 04:*:* 05:*:* 06:*:* 07:*:* 08:*:* 09:*:* 0a:*:* 0b:*:* 0d:*:* 0e:*:* 0f:*:* 10:*:* 11:*:* 12:*:* 13:*:* 14:*:* 3c:*:* dc:*:* e0:*:* ef:*:* fe:*:* ff:*:*}
+
+### Allow all mouses and keyboards, in a sense, so the user can conveniently change them without restrating the daemon.
```
ArrayBolt3, line 7: Typo fixes: `mouses` -> `mice`, `restrating` -> `restarting`

debian/control
```diff
@@ -30,6 +30,7 @@ Depends: adduser,
          python3,
          secure-delete,
          sudo,
+         usbguard,
```
ArrayBolt3, line 33: As mentioned previously, this needs to be removed from debian/control so that it can be shipped by a metapackage.

ArrayBolt3 (changes requested):
I like what this is intended to accomplish and think it's a great step forward. I'll give this some testing and report back how it works.

ArrayBolt3:
I tested this PR on physical hardware with the functional changes requested above. Surprisingly, my Logitech MX Mechanical Mini keyboard worked - upon closer examination it appears that it identifies itself as being a keyboard *and* a mouse that have boot interface support, and then *also* identifies itself as an HID device without boot interface support. Not sure how that works, but I'm guessing the keyboard and mouse bits were allowed while the "extra" device (probably for advanced configuration) was disallowed. A USB flash drive also worked just fine. I didn't test things that I expected should be rejected due to the limited number of USB ports in my system and devices available to me at the time.

ArrayBolt3:
I'll be integrating this PR (with the suggested changes) into my arraybolt3/trixie branch of security-misc. However, we absolutely *must* document how to loosen or remove these restrictions, since some laptops have internal keyboards and touchpads that connect via USB (the Framework 16 appears to fall into this category, see https://community.frame.work/t/keyboard-connection/32169/2). These settings will render those users unable to use external mice or keyboards with their laptops, which would be frustrating.

etc/usbguard/rules.d/30_security-misc.conf
ArrayBolt3: Weird... Red Hat's documentation actually has these same "invalid" rules... I wonder if people just don't follow the USB spec, or if Red Hat has something wrong...

PR #170, opened 2023-11-19
New line 990-security-misc.conf

monsieuremre:


adrelanos:
For hardened-malloc this is already here:
https://github.com/Kicksecure/hardened_malloc/blob/master/debian/hardened_malloc.conf

Is this security improving outside of hardened malloc? If so, please add some references.

monsieuremre:
My understanding is, probably. Because there already is ASLR. We randomize the address space layout. If there is more space, there are more possibilities for random layouts, so the layout is in fact more random and more secure. So if I'm not misinterpreting something, this should be a net positive.

adrelanos:
* https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/WVWHTLXSGZN4QMAE577ZFZX4ZI6YZF3A/
* https://www.suse.com/support/kb/doc/?id=000016692
* https://chat.openai.com/c/30ff8d53-6c6d-49d8-befe-05ebb9b7fb66

therealmate:
> https://chat.openai.com/c/30ff8d53-6c6d-49d8-befe-05ebb9b7fb66

broken link

monsieuremre:
So the consensus is larger=better but nor for security reasons, and rather for compatibility reasons. I still think having a larger max value is better for security too. Because when you think very simple: larger area=more random=more secure. Unfortunately there does not seems to be a lot of resources on if this would actually increase ASLR effectiveness. But, I don't see why that would not be the case.

adrelanos:
https://chat.openai.com/share/4fb17710-87ce-44ab-a187-d6eaba41e650

PR #171, opened 2023-11-19
Small systemd hardening

monsieuremre:
Just allow native architecture calls. From what I understand, 32 bit binaries would not benefit from ASLR due to the incredibly small address space and might cause other programs to have less effective ASLR too. Apparently this can also cause other security issues and it is in general recommended not to execute anything non-native.

Also once again core dump.

### DIFF
--- a/lib/systemd/system.conf.d/30_security-misc.conf
+++ b/lib/systemd/system.conf.d/30_security-misc.conf
@@ -1,0 +1,6 @@
+[Manager]
+DumpCore=no
+## Allow no non-native calls as these may be used to circumvent some security protections
+## For example 32 bit x86 code is executable in modern 64 bit x86 systems because of backwards
+## compatibility. This is undesirable and can potentially be dangerous.
+SystemCallArchitectures=native


lib/systemd/system.conf.d/30_security-misc.conf
```diff
@@ -0,0 +1,6 @@
+[Manager]
+DumpCore=no
```
adrelanos, line 2: 1) This can also be set by kernel command line. Better to set by config or append the already super long kernel command line?

2) Please add a link to the related systemd documentation.

3) Please add to readme.

4) Please undo in [debug-misc](https://github.com/Kicksecure/debug-misc) package and add to readme there.

These are all the steps that I would have to do otherwise. I might some day but could be considerably delayed until I get to that.

lib/systemd/system.conf.d/30_security-misc.conf
```diff
@@ -0,0 +1,6 @@
...
+## Allow no non-native calls as these may be used to circumvent some security protections
+## For example 32 bit x86 code is executable in modern 64 bit x86 systems because of backwards
+## compatibility. This is undesirable and can potentially be dangerous.
+SystemCallArchitectures=native
```
adrelanos, line 6: Seems high potential for breakage. Do I have specific examples? No. But I have an approach for such situations...

Could you please suggest changing the default upstream at systemd instead? That would probably result in valuable insights if this is useful, why it's not the upstream default and/or what breakage to expect.

monsieuremre: >Seems high potential for breakage. Do I have specific examples? No. But I have an approach for such situations...

Very big no. On a daily basis, you would rarely ever execute something of another architecture. For the most part, you can't anyway. For some, like 32 bit x86, you can do it because of backwards compatibility. But there aren't 32 bit x86 binaries on your 64 bit system anyway. This does not happen normally. x86 64 bit has been the industry standard for more than 20 years. Now, almost no operating system, even including the various linux distros, supports the 32 bit architecture, let alone the invidiual pieces of software. It bring no benefit to be able to execute those.

Other than that, there aren't any other use cases of supporting non native architectures, far as I know.

To be honest, since this never ever occurs, it also brings not too much extra security, because you won't make non native calls anyway. But it also brings no harm, IMO.

adrelanos: Ok. Please add to readme too.

PR #178, opened 2023-12-15
Disable asynchronous I/O

raja-grewal:
io_uring creation is disabled for all processes. io_uring_setup always fails with -EPERM. Existing io_uring instances can still be used.

https://security.googleblog.com/2023/06/learnings-from-kctf-vrps-42-linux.html

https://forums.whonix.org/t/io-uring-security-vulnerabilties/16890/4

### DIFF
--- a/usr/lib/sysctl.d/990-security-misc.conf
+++ b/usr/lib/sysctl.d/990-security-misc.conf
@@ -67,6 +67,9 @@ kernel.randomize_va_space=2
 kernel.unprivileged_bpf_disabled=1
 net.core.bpf_jit_harden=2
 
+## Disable asynchronous I/O for all processes.
+kernel.io_uring_disabled=2
+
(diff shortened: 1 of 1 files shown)


monsieuremre:
For the current debian LTS kernel this parameter seems to be invalid. As mentioned in the forum by you, we are not on 6.6 or a later version. How would this be beneficial on a debian system with an LTS  6.1 kernel?

adrelanos:
Compatibility with the kernel version in Debian stable is interesting but not a blocker for getting merged as long as nothing breaks.

Happy to merge this even if it does nothing. This is to avoid having too many pull requests lingering and too much burden contributors worrying about kernel versions.

Left this open for a while to entertain comments.

adrelanos:
* https://github.com/anthraxx/linux-hardened/issues/84

Issue #169, opened 2023-11-19
`Depends:` vs `Recommends:` vs none


adrelanos:
* `libpam-tmpdir`
  * https://github.com/Kicksecure/security-misc/pull/147
 * `usbguard`
   * https://github.com/Kicksecure/security-misc/pull/166

Do we really need `Depends:`? This makes it difficult to remove the package for users.

Perhaps `Recommends:` is sufficient? Not ideal.

Or nothing.

Why? Because security-misc shouldn't be attempting to be a Linux distribution such as Kicksecure that installs by a number of pre-installed applications.
  
references:

* https://github.com/Kicksecure/kicksecure-meta-packages
* https://github.com/Kicksecure/kicksecure-meta-packages/debian/control

monsieuremre:
I am for depends.

Recommends is too weak and we configure the config files. Security-misc does not try to be a distribution in my opinion. If it becomes a vital component of kicksecure, that is still ok. I see no problem. I also think these packages are relatively minimalistic.

adrelanos:
Depends in security-misc would result in usbguard being installed by default in Kicksecure for VirtualBox VM images and these don't come with USB enabled by default so that would be a superfluous package. Everybody without USB couldn't remove this package without getting trouble with security-misc.

The usbguard config suggestion seems suitable here but the depends can go to kicksecure-meta-packages.

Maybe in

* kicksecure-recommended-cli (libpam-tmpdir, usbguard) or
* non-qubes-vm-enhancements-cli (usbguard)

not sure yet.

adrelanos:
`Depends:` would be inappropriate as per Debian policy.

Packaging is done with Debian policy compliance in mind. Packages by Kicksecure, Whonix are already `lintian --pedantic` clean as much as feasible.

monsieuremre:
At least recommend them. I don't support moving them to another package. Everything that is security-misc should be handled here. This provides platform indepence. It shoud be kept possible to install this on debian and have a hardened system.

adrelanos:
I exactly don't want to replicate Kicksecure in security-misc. There's a structure.

If starting to add security enhancing recommended packages which then would become candidates for `Recommends:`.

* LKRG
* tirdad
* amd64-microcode / intel-microcode
* swap-file-creator (encrypted swap file)
* ... others, see kicksecure-meta-packages debian/control

[...]

Issue #168, opened 2023-11-18
My Argument for KDE | Developers in the forum disregarded the idea too quickly


monsieuremre:
As the title says. I think whonix maintainers in the forum have discarded the idea of moving back to kde too quickly. Moreover, I find most all of the arguments to be for the most part unfounded. KDE being bloated or heavy for Whonix is just not true. KDE used to be 'heavier' when both Qubes and Whonix were using it. It is now much more optimized. Xorg is unironically massively more bloated than wayland. All the 'bloat' is not part of the DE. We do not have to install everything. Just install the shell and we are done.

And I want to make this clear, I propose we only install the shell. Not even the terminal, not the file manager, nothing from the debian repositories but the shell, and let me explain why. But before that, I want to recap once again why xfce severly slows whonix down.

* Xfce is now officially the only 'major' DE that does not have a plan to switch to wayland for good. They say they would try stuff, but that's it. It is not even certain that they will switch at all. Even Cinnamon has given a date now. Budgie, Mate, they are all switching, they have plans, they have dates. Stuff like Lxqt are already almost completely switched. Xfce is an outlier here. And this is expected. They for some reason agreed not to use wlroots or libmutter. That means they kinda have to develop their own compositor, which we all know is not possible with the little funding Xfce has. If they do not decide to suddenly base on wlroots, Xfce will never have wayland.
* Xfce is lags behind, massively. On everything. The desktop looks really, really, but I mean really outdated. Normally this is not as important but you have to keep in mind, compelling to the average user is something you probably want.
* It is really a shame whonix still unironically ships with xorg as default.
* The only reason this is still just passable is whonix designed to be run on a VM. But Kicksecure is planned to be a regular distro. That means, xorg gotta go, really. True and real.
* Xfce is small. You tell on the Whonix website that you base on Debian, and one of the reasons for that is that there is a good chance in 10 years Debian is still going to be around. Do you know what not might be around in 10 years? Xfce. And by 'might', I mean they won't definetely be around, if they don't catch the pace in switching to wayland.

So at the very least, before I make the argument for KDE, we should all more or less agree on this: Xfce bad. At least for security.

Now my argument for KDE.
* KDE e.V. is a well-established and well-funded non-profit organization based in Germany. They are not some random project.
* KDE is customizable. It is fast. It is optimized. It can be as simple and minimalistic, if not more minimalistic, than xfce on demand.
* KDE has a well functioning wayland compositor. Which is to be the default.
[...]

PR #193, opened 2024-01-24
Hardened Malloc

monsieuremre:
Depends on hardened malloc and enables it for all system and user services as default. User applications are thus unaffected, like web browsers etc. Most sensitive and vulnerable OS components use hardened malloc and are hardened against memory corruption and heap exploits.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -29,6 +29,7 @@ Depends: adduser,
          python3,
          secure-delete,
          sudo,
+         hardened-malloc,
          ${misc:Depends}
 Replaces: anon-gpg-tweaks, swappiness-lowest, tcp-timestamps-disable
 Description: Enhances Miscellaneous Security Settings
--- a/usr/lib/systemd/system.conf.d/50_security-misc.conf
+++ b/usr/lib/systemd/system.conf.d/50_security-misc.conf
@@ -1,0 +1,4 @@
+[Manager]
+DefaultEnvironment="LD_PRELOAD='libhardened_malloc.so'"
+SystemCallArchitectures=native
+DumpCore=no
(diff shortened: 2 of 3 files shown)


adrelanos:
Thank you for your pull request! Unfortunately, this is the wrong repository.

security-misc is the wrong package for this sort of `Depends:`. That belongs into `kicksecure-meta-packages`. As per:
https://github.com/Kicksecure/security-misc/issues/169

https://github.com/Kicksecure/kicksecure-meta-packages

Enabling hardened malloc by default belongs for system and user systemd units the `hardened-malloc`. 

https://github.com/Kicksecure/hardened_malloc

That source code repository currently generates two packages:

* `hardened-malloc`
* `hardened-malloc-light-enable`

Welcome contributions:

* A) A contribution for `hardened-malloc-default-enable`. Or,
* B) Enable hardened malloc default in package `hardened-malloc` by default for user and systemd units.

For B) merging would be blocked if it breaks X11 inside VirtualBox. That would be blocked until port to Wayland is complete. Related:

* https://github.com/Kicksecure/security-misc/issues/168
* https://forums.whonix.org/t/port-to-wayland/17380

Other blockers would be if browsers are broken by default. This is because enabling hardened malloc for systemd units might get inherited by the DE (desktop environment) all the way until the browser.

monsieuremre:
This would only affect services. If anything is affected, it can be exempted easily with the same manner. This may not be the right place, you are right. Tho there is no way to open an issue here:

https://github.com/Kicksecure/hardened_malloc

adrelanos:
> This may not be the right place, you are right. Tho there is no way to open an issue here:
> 
> https://github.com/Kicksecure/hardened_malloc

Fixed.

Issue #185, opened 2024-01-06
Restrict umask to 027 except for sudo/root broken


adrelanos:
Currently umask is set to `027` (read, write for owner and group only).
(Group is OK because Debian uses `usergroups` by default, [`UPG`](https://wiki.debian.org/UserPrivateGroups) (UserPrivateGroups)).

This however should not be the case for files created by root as this is super confusing, causing issues when creating files in `/etc` (or `/usr`) as these won't be readable by applications normally suppose to be able to read these running as non-root. That part is broken.

Therefore unfortunately this feature has to be reverted until this can be fixed.

This was discussed before here:

* https://github.com/Kicksecure/security-misc/pull/18

The following unfortunately has to be removed.

[`/usr/share/pam-configs/umask-security-misc`](https://github.com/Kicksecure/security-misc/blob/master/usr/share/pam-configs/umask-security-misc)

```
Name: Restrict umask to 027 (by package security-misc)
Default: yes
Priority: 100
Session-Type: Additional
Session-Interactive-Only: yes
Session:
	[success=1 default=ignore]	pam_succeed_if.so uid eq 0
	optional	pam_umask.so umask=027
```

I cannot even just out-comment it to make it easier to tinker as folder `/usr/share/pam-configs` does not support comments. Populating `/etc/pam.d` from `/usr/share/pam-configs` is a Debian / Ubunut feature.

* https://unix.stackexchange.com/questions/337808/where-is-the-official-documentation-for-the-files-in-usr-share-pam-configs-in
* https://wiki.ubuntu.com/PAMConfigFrameworkSpec

----

Command for testing:

    touch a && ls -la a && rm a

I tried various stuff to fix it:

* delete `Session-Interactive-Only: yes`
*

```
        [success=2 default=ignore]      pam_succeed_if.so debug uid eq 0
        [success=1 default=ignore]      pam_succeed_if.so debug use_uid uid eq 0
	optional	pam_umask.so debug umask=027
```

----

This issue is only happening in non-Qubes. Not reproducible in Qubes Debian. This might be because Qubes login session work differently. This is probably also why I did not spot this issue beforehand.

----

Contributing: I don't necessarily need the config snippet for `/usr/share/pam-configs` as I might be able to figure that out but the config snipped required for `/etc/pam.d` would very useful.

adrelanos:
Disabled just now in git.

monsieuremre:
What happens when you run the command ```umask``` with the old config:
* When directly root
[...]

PR #194, opened 2024-01-24
Umask Settings

monsieuremre:
Set umask as 0077 for all users and services using the pam module. Then override this by setting the umask 0022 for ```root``` under ```/etc/environment.d/```. Any valid use case where having a umask of 0077 can be problematic is thus covered and not included in the hardening. System administrators will have this umask when they login.

### DIFF
--- a/etc/profile.d/30_security-misc-umask.sh
+++ b/etc/profile.d/30_security-misc-umask.sh
@@ -1,0 +1,6 @@
+if [ "$(id -u)" -eq 0 ]
+then
+    umask 077
+else
+    umask 022
+fi
--- a/usr/share/pam-configs/mkhomedir-security-misc
+++ b/usr/share/pam-configs/mkhomedir-security-misc
@@ -4,4 +4,4 @@ Priority: 100
 Session-Type: Additional
 Session-Interactive-Only: yes
 Session:
-	optional	pam_mkhomedir.so umask=027
+	optional	pam_mkhomedir.so umask=0077
--- a/usr/share/pam-configs/umask-security-misc
+++ b/usr/share/pam-configs/umask-security-misc
@@ -1,0 +1,8 @@
+Name: Restrict umask to 0077 (by package security-misc)
+Default: yes
+Priority: 100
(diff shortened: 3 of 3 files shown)


adrelanos:
Needs at minimum this:
https://github.com/Kicksecure/security-misc/blob/master/debian/security-misc.links
?

This needs some test cases for the different environments.
Could you please create a list of different environments (mentions in https://github.com/Kicksecure/security-misc/issues/185), add to https://www.kicksecure.com/wiki/Dev/Strong_Linux_User_Account_Isolation and test what you can?

monsieuremre:
> Could you please create a list of different environments (mentions in https://github.com/Kicksecure/security-misc/issues/185), add to https://www.kicksecure.com/wiki/Dev/Strong_Linux_User_Account_Isolation and test what you can?

What do you mean? We can also set the umask for ssh too, if we want some different value there. But I don't get what is meant by different environments.

adrelanos:
> What do you mean?

- sudo su
- su
- ssh login as user
- ssh login as root
- login on virtual terminal as user
- - login on virtual terminal as root
- X11 login
- Wayland login
- Qubes dom0 sudo xl console vmname
- ... as mentioned in that ticket

monsieuremre:
Oh ok. Su sudo and manual execution of root privileged commands are covered with the exemption. So they will use the weak umask. Logins are also covered obviously. SSH logins are not covered, but that can be set separately as I said.

When it comes to X11 and Wayland login, I have no idea. If these run as services or daemons, then they are not exempted. But here we can use systemd drop in files if necessary. But I don't see any downsides to these things having strong umasks.

Qubes dom0 sudo xl console vmname - no idea. Qubes is very different, cannot know without testing.

adrelanos:
I need a checklist before this can proceed further. (No need to run all 
tests by yourself.)

One would be created by me one day but that could take time. Faster if 
contributed.

adrelanos:
Stalled.

Superseded by:

* https://github.com/Kicksecure/security-misc/pull/282

Issue #172, opened 2023-11-19
improve hide-hardware-info.service, make `/sys` hiding optional


monsieuremre:
```
A systemd service restricts /proc/cpuinfo, /proc/bus, /proc/scsi and /sys to the root user. This hides a lot of hardware identifiers from unprivileged users and increases security as /sys exposes a lot of information that shouldn't be accessible to unprivileged users. As this will break many things, it is disabled by default and can optionally be enabled by executing systemctl enable hide-hardware-info.service as root.
```
Ok.

Although not tested, I am almost 100% sure practically all the breakage is caused by restricting ```/sys```. My recommendation is, we leave this one out. Add ```/proc/kallsyms``` to the list as well, as this contains all the memory addresses for each kernel symbol. And enable what we already have, restricting the sub-directories under proc.

Should I follow with the respective pull?

adrelanos:
I am not sure how that PR would look but I guess rather small.

Since this is not enabled by default and for extra security, I tend to add what you suggested here.

Worth trying.

monsieuremre:
Actually we do not even need that service I think. We can set the permissions using the permission hardener service. We are just going to do these with the already existing permission hardener service.

chmod 400 /proc/kallsyms
chmod 400 /proc/cpuinfo
chmod 500 /proc/bus

And that's it. This hides a lot of hardware and kernel info. And I am pretty positive it won't break anything.

adrelanos:
`hide-hardware-info.service` issues:

* [broken flatpak example (`flatpak run org.chromium.Chromium`)](https://forums.whonix.org/t/restrict-hardware-information-to-root-testers-wanted/8618/51)
* [Xfce restart and shutdown button](https://forums.whonix.org/t/restrict-hardware-information-to-root-testers-wanted/8618/55)

But these are all probably related to `/sys`.

Could you test please if these work?

related:
https://www.kicksecure.com/wiki/Security-misc#Restrict_Hardware_Information_to_Root

adrelanos:
> Actually we do not even need that service I think. We can set the permissions using the permission hardener service. We are just going to do these with the already existing permission hardener service.

That's an interesting idea to get rid of the extra service. Nice in theory, but...

Could you check please what other features the current implementation has?

* https://github.com/Kicksecure/security-misc/blob/master/lib/systemd/system/hide-hardware-info.service
* https://github.com/Kicksecure/security-misc/blob/master/usr/libexec/security-misc/hide-hardware-info

[...]

PR #202, opened 2024-02-20
Secure Bind Mount & Protect Hardware

monsieuremre:
Hello. This pull request resolves two currently open issues. One is protecting hardware by limiting access to /proc and /sys. Well good news. I present you the best way possible. Mounting proc with subset=pid ensure that only the pid subset under proc is visible, everything else is invisible. This is also the case for root. No one can see anything. ```/proc/dma```, ```/proc/iomem```, ```/proc/ioports```, ```/proc/meminfo```, ```/proc/kallsyms```, ```/proc/kcore``` and literally tons of other stuff, everything protected even from root. Just ```ls /proc``` to see for yourself what gets hidden.

When it comes to /sys, [starting based on this comment](https://github.com/Kicksecure/security-misc/issues/172#issuecomment-1951419805), I found the sources of breakages. The comment is true, but not complete, because there are also other subsets that break too, apart from just fs. Those are enabled, and everything else is disabled. Tested.

And bind mounting. You said not having bind mounts is a big deal. You said binds are important, they are our lives. Well. You say no more. I give you binds, a lot of them, and them only. At this point, we might as well just bind everything, disregarding if it has a proper real partition. Because even if it does have a partition, bind mounting to itself does no harm literally. The overhead seems to be really really minimal and unnoticable. So this is the current suggestion. Test for yourself and report please.

### DIFF
--- a/usr/bin/secure-bind-protect-hardware
+++ b/usr/bin/secure-bind-protect-hardware
@@ -1,0 +1,30 @@
+#!/bin/bash
+mount --options defaults,nosuid,nodev --bind --make-private /opt /opt
+mount --options defaults,nosuid,nodev --bind --make-private /home /home
+mount --options defaults,nodev,nosuid --bind --make-private /tmp /tmp
+mount --options defaults,nodev,nosuid,noexec --bind --make-private /boot/efi /boot/efi
+mount --options defaults,nodev,noosuid,noexec --bind --make-private /boot /boot
+mount --options defaults,nodev --bind --make-private /usr /usr
+mount --options defaults,nosuid,nodev --bind --make-private /var /var
+mount --options defaults,nosuid,nodev,noexec --bind --make-private /etc /etc
+mount --options defaults,nosuid,nodev,noexec --bind --make-private /srv /srv
+mount --options defaults,nosuid,nodev,noexec --bind --make-private /root /root
+mount --options defaults,nosuid,nodev,noexec --bind --make-private /media /media
+mount --options defaults,nosuid,nodev --bind --make-private /mnt /mnt
+mount --options defaults,nosuid,nodev,noexec,remount,subset=pid /proc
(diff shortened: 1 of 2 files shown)


adrelanos:
Thank you, there's some really nice research here!

I am afraid, we are going in cycles here. Some stuff disregarded here...

If a systemd unit file based implementation, then there's already a better systemd unit file here:
https://github.com/Kicksecure/security-misc/blob/master/usr/lib/systemd/system/remount-secure.service

related:
https://www.kicksecure.com/wiki/Dev/remount-secure#systemd_unit_file_based_mount_options_hardening

The existing script `remount-secure` seems also better:
https://github.com/Kicksecure/security-misc/blob/master/usr/bin/remount-secure

* configureable
* error handling
* log output

I would prefer to improve that instead.

* Any wrong / missing mount options (if any).
* `--make-private` (which seems interesting)

Secure mount options should not be mixed with:

* https://github.com/Kicksecure/security-misc/issues/172

I am not sure this can be merged. But just leave it open. There are really good suggestions here.

usr/bin/secure-bind-protect-hardware
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+mount --options defaults,nosuid,nodev --bind --make-private /opt /opt
+mount --options defaults,nosuid,nodev --bind --make-private /home /home
+mount --options defaults,nodev,nosuid --bind --make-private /tmp /tmp
+mount --options defaults,nodev,nosuid,noexec --bind --make-private /boot/efi /boot/efi
+mount --options defaults,nodev,noosuid,noexec --bind --make-private /boot /boot
```
adrelanos, line 6: Type `noosuid` here but nvm.

adrelanos:
Changes suggested here:

* remount related: made it into remount-secure script and systemd unit.
* into the related issue https://github.com/Kicksecure/security-misc/issues/172

PR #204, opened 2024-02-22
Make /sys hardening optional and allow access to /sys/fs to make polkit work

### LINKED ISSUE #172: improve hide-hardware-info.service, make `/sys` hiding optional

DanWin:
Fixes #172

### DIFF
--- a/etc/hide-hardware-info.d/30_default.conf
+++ b/etc/hide-hardware-info.d/30_default.conf
@@ -7,6 +7,9 @@
 ## Disable the /proc/cpuinfo whitelist.
 #cpuinfo_whitelist=0
 
+## Disable /sys hardening.
+#sysfs=0
+
 ## Disable selinux mode.
 ## https://www.whonix.org/wiki/Security-misc#selinux
 #selinux=0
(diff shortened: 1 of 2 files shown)


monsieuremre:
Have this been tested? The reason I had more subsets of /sys as exceptions is that I experienced breakage with them hidden. /sys/fs for polkit was not enough. A difference of DE's maybe? Seems rushed to be enabled by default. Will comment if I find the time to test again and to report if breakage happens.

adrelanos:
Tested.

hide-hardware-information.service hasn't been enabled by default.

Not enabled by default as per:
https://github.com/Kicksecure/security-misc/blob/master/usr/lib/systemd/system-preset/50-security-misc.preset#L5

There's no ticket to enable it by default either.

PR #218, opened 2024-05-01
More CPU Mitigations and Additional References

raja-grewal:


adrelanos:
`gather_data_sampling=force` was enabled here to Enable Gather Data Sampling (GDS) mitigation. This is related to CPU `AVX` instruction.

adrelanos:
This has broken:

* `wc -l` (fixed in Debian `trixie`)
* `rspamd --version` (rspamd totally broken) (reported upstream)

Error message:

> illegal hardware instruction (core dumped)

More information here:

* https://github.com/rspamd/rspamd/issues/5137

PR #233, opened 2024-07-14
Refactor `/etc/default/grub.d/*`

raja-grewal:
Refactor existing grub kernel parameters for clarity, ease-of-use, and future-proofing.

Add two (deactivated) kernel parameters to further reduce attack surface.

## Changes

Merge `40_distrust_bootloader.cfg`, `40_distrust_cpu.cfg`, `40_enable_iommu.cfg`, and `40_kernel_hardening.cfg` into a single organised file `40_kernel_hardening.cfg`.

Rename `40_only_allow_signed_modules.cfg` to `40_signed_modules.cfg`.

Rename `40_remmount-secure.cfg` to `40_remount_secure.cfg`.

Rename `41_quiet.cfg` to `41_quiet_boot.cfg`.

There are no changes to the actual functionality of the code.

## Mandatory Checklist

## Optional Checklist


adrelanos:
If renaming or deleting a file under `/etc` these need to be added to https://github.com/Kicksecure/security-misc/blob/master/debian/security-misc.maintscript - otherwise, these will not be removed for users that upgrade this package.

etc/default/grub.d/40_kernel_hardening.cfg
```diff
@@ -5,58 +5,196 @@ kpkg="linux-image-$(dpkg --print-architecture)" || true
...
+#ia32_emulation=0
+
+## 2. Direct Memory Access:
+##
+## https://madaidans-insecurities.github.io/guides/linux-hardening.html#dma-attacks
+
+## Enable CPU manufacturer-specific IOMMU drivers to protect against DMA attacks.
+##
+GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX amd_iommu=force_isolation"
+GRUB_CMDLINE_LINUX="$GRUB_CMDLINE_LINUX intel_iommu"
```
adrelanos, line 131: This is a bug. `intel_iommu` without `=` is invalid. Before this was `intel_iommu=on`.

adrelanos:
This is a bit difficult to review that only comments were changed and no functionality was changed. The diff looks difficult to review. Found 1 bug above.

```
cd etc/default/grub.d
```

```
grep -r --invert-match '^#\|^$'
```

(Add `-h` to hide `grep` prepending the filename.)

Repeat the same command for the other branch. Then compare the two outputs with a diff viewer and make sure there are only expected changes as in splitting two options into multiple lines.

Could you please recheck using this or similar methodology that nothing was changed?

adrelanos:
Also helpful might be to append `| sort`.

    grep -h -r --invert-match '^#\|^$' | sort

adrelanos:
I am not sure about https://github.com/Kicksecure/security-misc/pull/233/commits/f4d652fa7b5dd350b577521c6bba22c9eb3c13f1.

Some (Debian?) default configuration file adds `quiet` by default. (Or has this been changed?) If we don't remove that first `quiet` but only append our own `quiet` this will result in the kernel command line having `quiet` twice. Hence the idea of

```
GRUB_CMDLINE_LINUX_DEFAULT="$(echo "$GRUB_CMDLINE_LINUX_DEFAULT" | LANG=C str_replace "quiet" "")"
```

was to remove the default `quiet` and add a new one which then will be easier for the user to locate and disable.

PR #242, opened 2024-07-18
Disable the usage of `ptrace()` by all processes

raja-grewal:
Upgrade existing `sysctl` to now disable `ptrace()` usage by all processes.

Previously it was restricted to users with `CAP_SYS_PTRACE`.

It is unlikely that this would cause significantly more software breakages than the previous setting.

Needs to be thoroughly tested before being upgraded.

## Changes

`kernel.yama.ptrace_scope=2`
to
`kernel.yama.ptrace_scope=3`

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,8 @@ space, user space, core dumps, and swap space.
 
 - Disable asynchronous I/O (when using Linux kernel version >= 6.6).
 
-- Restrict usage of `ptrace()` to only processes with `CAP_SYS_PTRACE` as it
-  enables programs to inspect and modify other active processes. Provide the
-  option to entirely disable the use of `ptrace()` for all processes.
+- Disable the usage of `ptrace()` by all processes as it enables programs to
+  inspect and modify other active processes.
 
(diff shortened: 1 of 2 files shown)


adrelanos:
Being restricted to `CAP_SYS_PTRACE` seems sufficient?

adrelanos:
> It probably is for most purposes but one flaw in the current approach is that it can be changed through `sysctl` on a running environment.

But that's okay because that requires root and for now, as soon as root is compromised it's considered game over anyhow?

adrelanos:
There are probably a ton of settings which are unsafe to change.

If there was an option to lock all settings until reboot (I think there is not), should we do that?

User education by locking settings and requiring reboot seems out-of-scope of security-misc.

Related:

* https://www.kicksecure.com/wiki/Dev/boot_modes
  * Because that introduces concepts such as secureadmin (untrusted root) and superadmin. If that was implemented, then reboot would be required indeed to change any kind of settings.

PR #248, opened 2024-07-24
Re-enable (default) `secure_redirects` for ICMP redirect messages

raja-grewal:
Previously going back even further than  [4 years](https://github.com/Kicksecure/security-misc/blob/6a4c493213929b354a3c8d2acf2325473ae63cfd/etc/sysctl.d/30_security-misc.conf) we have actually not enabled secure redirects for ICMP redirect messages over IPv4.

This error is also in Madaidan's [guide](https://madaidans-insecurities.github.io/guides/linux-hardening.html#sysctl-network).

However, this mistake has not resulted in any breakages since we do not accept ICMP redirects by default:
```
net.ipv4.conf.*.accept_redirects=0
net.ipv4.conf.*.send_redirects=0
net.ipv6.conf.*.accept_redirects=0
```

If these were to be commented-out and restored back to `=1`'s, we would actually accept ICMP redirects for ALL gateways which is wrong!

See:
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/security_guide/sect-security_guide-server_security-disable-source-routing#sect-Security_Guide-Server_Security-Disable-Source-Routing
https://www.frozentux.net/ipsysctl-tutorial/chunkyhtml/theconfvariables.html
https://www.debian.org/doc/manuals/securing-debian-manual/network-secure.en.html
https://askubuntu.com/questions/118273/what-are-icmp-redirects-and-should-they-be-blocked

## Changes

`net.ipv4.conf.*.secure_redirects=0`
to
`net.ipv4.conf.*.secure_redirects=1` 

But `net.ipv4.conf.*.secure_redirects=1` is the default and so we should actually just remove the incorrect `systcl` as they are redundant.

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,8 +72,7 @@ Various networking components of the TCP/IP stack are hardened for IPv4/6.
   from all interfaces to prevent IP spoofing.
 
 - Disable ICMP redirect acceptance and redirect sending messages to
-  prevent man-in-the-middle attacks and minimize information disclosure. If
-  ICMP redirect messages are permitted, only do so from approved gateways.
+  prevent man-in-the-middle attacks and minimize information disclosure.
 
 - Ignore ICMP echo requests to prevent clock fingerprinting and Smurf attacks.
 
--- a/README.md
+++ b/README.md
@@ -72,7 +72,8 @@ Various networking components of the TCP/IP stack are hardened for IPv4/6.
   from all interfaces to prevent IP spoofing.
 
 - Disable ICMP redirect acceptance and redirect sending messages to
-  prevent man-in-the-middle attacks and minimize information disclosure.
+  prevent man-in-the-middle attacks and minimize information disclosure. If
+  ICMP redirect messages are permitted, only do so from approved gateways.
 
 - Ignore ICMP echo requests to prevent clock fingerprinting and Smurf attacks.
 
--- a/README.md
+++ b/README.md
@@ -71,9 +71,8 @@ Various networking components of the TCP/IP stack are hardened for IPv4/6.
 - Enable reverse path filtering (source validation) of packets received
   from all interfaces to prevent IP spoofing.
 
-- Disable ICMP redirect acceptance and redirect sending messages to
-  prevent man-in-the-middle attacks and minimize information disclosure. If
-  ICMP redirect messages are permitted, only do so from approved gateways.
+- Disable ICMP redirect acceptance and redirect sending messages to prevent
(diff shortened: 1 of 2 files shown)


adrelanos:
This is confusing me a bit. So if we out comment the secure_redirects then it's still fine because all redirects are disabled.

If a user then would out-comment the disabling for all redirects, i.e. re-enable redirects, the user would end up with secure redirects only, which is better than accepting all redirects?

Or maybe best is to keep disabling "all redirects" (all, default, secure) but move these options closer to each other and document this better so if a user was to re-enable redirects the user would be informed about all vs "secure" only?

adrelanos:
Since it's the default, best to comment these out and mention it's the default?

Also this default is unlikely to ever change.

And then explain all of this.

By moving closer together I mean changing from:


```## Disable ICMP redirect acceptance and redirect sending messages.
## Prevents man-in-the-middle attacks and minimizes information disclosure.
##
## https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/6/html/security_guide/sect-security_guide-server_security-disable-source-routing#sect-Security_Guide-Server_Security-Disable-Source-Routing
## https://www.frozentux.net/ipsysctl-tutorial/chunkyhtml/theconfvariables.html
## https://www.debian.org/doc/manuals/securing-debian-manual/network-secure.en.html
## https://askubuntu.com/questions/118273/what-are-icmp-redirects-and-should-they-be-blocked
##
net.ipv4.conf.all.accept_redirects=0
net.ipv4.conf.default.accept_redirects=0
net.ipv4.conf.all.send_redirects=0
net.ipv4.conf.default.send_redirects=0
net.ipv6.conf.all.accept_redirects=0
net.ipv6.conf.default.accept_redirects=0

## Accept ICMP redirect messages only for approved gateways.
## If ICMP redirect messages are permitted, only useful if managing a default gateway list.
##
## https://github.com/Kicksecure/security-misc/pull/248
##
net.ipv4.conf.all.secure_redirects=1
net.ipv4.conf.default.secure_redirects=1
```

To:

```net.ipv4.conf.all.accept_redirects=0
net.ipv4.conf.default.accept_redirects=0
net.ipv4.conf.all.send_redirects=0
net.ipv4.conf.default.send_redirects=0
net.ipv6.conf.all.accept_redirects=0
net.ipv6.conf.default.accept_redirects=0
net.ipv4.conf.all.secure_redirects=1
net.ipv4.conf.default.secure_redirects=1
```

PR #249, opened 2024-07-25
Disallow registering interpreters for miscellaneous binary formats

raja-grewal:
Disallow registering interpreters for various (miscellaneous) binary formats based on a magic number or their file extension to prevent unintended code execution.

Breaks many scripts that do not have appropriate shebang interpreter directives (#!/bin/...).

## Changes

Set `sysctl fs.binfmt_misc.status=0`.

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,8 +57,9 @@ space, user space, core dumps, and swap space.
 
 - Increase the maximum number of memory map areas a process is able to utilize.
 
-- Disallow registering interpreters for various (miscellaneous) binary formats based
-  on a magic number or their file extension to prevent unintended code execution.
+- Provide the option to disallow registering interpreters for various (miscellaneous)
+  binary formats based on a magic number or their file extension to prevent
+  unintended code execution.
 
 - Disable core dump files and prevent their creation. If core dump files are
   enabled, they will be named based on `core.PID` instead of the default `core`.
--- a/README.md
+++ b/README.md
@@ -57,9 +57,8 @@ space, user space, core dumps, and swap space.
 
 - Increase the maximum number of memory map areas a process is able to utilize.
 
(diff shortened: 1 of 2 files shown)


adrelanos:
But for execution, the execution bit ("`chmod +x`") is still required? No execution bit -> no binfmt_misc -> no security issue?

The execution bit afaik cannot set itself.

Also trivial to circumvent for an attacker. Instead of only using `.sh` the attack would keep using `.sh` but also add a proper shebang? So this, while interesting, I don't see what advantage it provides?

adrelanos:
> It might still be useful perhaps against older existing legacy malware that has yet to adapt?

Not sure there is any legacy malware where this would have been the case?

>     1. Close it since it is likely not that effective against modern attack vectors.

Yes.

>     2. Comment out the `sysctl` but leave the documentation there for future reference with a link to this PR.

This seems useful.

adrelanos:
Yes. Could you please document with this reasoning?

adrelanos:
Looks great!

Where can we see that `fs.binfmt_misc.status` is a valid kernel setting? I cannot find it on kernel.org. Search term:

> fs.binfmt_misc.status site:kernel .org

Cannot find a manual for it.

    sudo sysctl -a | grep fs.binfmt_misc.status

> fs.binfmt_misc.status = enabled

That would imply that we want to set `fs.binfmt_misc.status=disabled`. 

How to actually test if the setting change takes effect?

adrelanos:
Alright, `sudo sysctl -w fs.binfmt_misc.status=1` vs `0` switches from `enabled` to `disabled` when checking using `sudo sysctl fs.binfmt_misc.status` so this seems correct indeed.

PR #252, opened 2024-07-31
Mitigate tar storing usernames and groups

groovy-boiler:
tar stores your user and its group in the archive.
Make an archive with tar (or find one) then extract it like this:

`tar -zvtf archive-name.tar.gz | head -n 10`

The output will look like something like this:

```
-rw-r--r-- kevin/volunteer    290 2023-09-11 03:10 file1
-rw-r--r-- kevin/volunteer  23558 2024-07-30 11:45 file2
-rw-r--r-- bob/admin        10958 2024-07-30 12:54 file3
-rw-r--r-- frank/staff        110 2023-07-04 11:45 file4
```

`--owner="user" --group="user"` will mitigate this regardless of what the user logged in as is when creating a archive with tar.
## Mandatory Checklist

- [X] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

## Optional Checklist

- [X] I have tested it locally
- [ ] I have reviewed and updated any documentation if relevant

Fixes #
**Privacy by default when executing tar to make an archive**

### DIFF
--- a/usr/local/bin/tar
+++ b/usr/local/bin/tar
@@ -1,0 +1,5 @@
+#!/bin/sh
+## tar stores usernames and groups
+## mitigate showing user/group in metadata
+## set user and group to user/user
+exec /bin/tar --owner="user" --group="user" "$@"


adrelanos:
This kind of privacy seems out-of-scope for Kicksecure. What is in-scope 
and out-of-scope is elaborated here:

https://www.kicksecure.com/wiki/Privacy

groovy-boiler:
Well to be fair this is kinda a security thing though...

Kicksecure does set the default user to "user" but someone might not want to use user or have other users. I don't see how this would hurt having this in security-misc to be honest?

How about create something like a *privacy-misc* with things like this in it?

groovy-boiler:
sad ngl

PR #261, opened 2024-08-16
Simplify syntax of some network-related `sysctl`'s

raja-grewal:
This pull request simplifies the syntax of some network-related `sysctl`'s.

I do not think there is a need to distinguish between application to  'all' and 'default'. Ideally these setting should be applied across the board regardless of interface.

One could also make the case this might be a form of weak 'hardening' as we have simplified each setting to one line (defence-in-depth etc.).

Note this approach is also currently used by GrapheneOS's [infrastructure](https://github.com/GrapheneOS/infrastructure/blob/main/sysctl.d/local.conf).

For example:
```
net.ipv4.conf.all.accept_redirects=0
net.ipv4.conf.default.accept_redirects=0
```
to
```
net.ipv4.conf.*.accept_redirects=0
```

## Changes

There are (likely) no changes to the  functionality of the code.

EDIT: This PR attempts to fix a bug in the existing `rp_filter` implementation, see below. 

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/usr/lib/sysctl.d/990-security-misc.conf
+++ b/usr/lib/sysctl.d/990-security-misc.conf
@@ -275,8 +275,7 @@ net.ipv4.tcp_rfc1337=1
 ## https://forums.whonix.org/t/enable-reverse-path-filtering/8594
 ## https://seclists.org/oss-sec/2019/q4/122
 ##
-net.ipv4.conf.all.rp_filter=1
-net.ipv4.conf.default.rp_filter=1
+net.ipv4.conf.*.rp_filter=1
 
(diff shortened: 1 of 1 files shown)


adrelanos:
Could you test please (as root [1])...

1 Record old.

    sysctl -a > old

2 Use the new settings.

3 Apply the new settings.

    sysctl --system

4 Record new.

    sysctl -b > new

5 Compare (with favorite diff viewer).

    diff old new

Does it result in effectively the same sysctl settings as understood by the kernel or will there be a difference?

Also would be good to document this test somehow so we can refer to it for sysctl refactoring in the future.

----

[1] Obvious to you but not all readers.

PR #264, opened 2024-08-18
Add KSPP compliance notices to corresponding parameters and `sysctls`

raja-grewal:
This PR adds KSPP compliance notices to corresponding parameters and `sysctl`s as per https://github.com/Kicksecure/security-misc/issues/256.

I have tried to be comprehensive and thorough but please re-check my work.

For the time being I have decided not to add "KSPP=no" notices as I am not sure that these would be helpful? Is there any point in showing non-compliance if we already show what is consistent with the KSPP.

Also there are two areas of of partial compliance:

```
## KSPP sets the stricter sysctl user.max_user_namespaces=0.
##
kernel.unprivileged_userns_clone=0

## KSPP sets the stricter sysctl kernel.yama.ptrace_scope=3.
##
kernel.yama.ptrace_scope=2
```

Do we want to adhere to the KSPP on these?

The first is discussed in https://github.com/Kicksecure/security-misc/pull/263 and has the potential to cause breakages but I am not that familiar with its real world usage.

The second was discussed earlier in https://github.com/Kicksecure/security-misc/pull/242.

## Changes

There are currently no changes to the functionality of the code.

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #268, opened 2024-08-28
Enable `panic_on_warn=1`

raja-grewal:
Partial follow-up on https://github.com/Kicksecure/security-misc/pull/264#issuecomment-2314954119.

This pull request sets `sysctl kernel.panic_on_warn=1` as per the KSPP [recommendations](https://kspp.github.io/Recommended_Settings#sysctls).

```
# Reboot after even 1 WARN or BUG/Oops. Adjust for your tolerances. (Since v6.2)
# If you want to set oops_limit greater than one, you will need to disable CONFIG_PANIC_ON_OOPS.
kernel.warn_limit = 1
kernel.oops_limit = 1
```

The proposed setting is not exactly equivalent as the KSPP assumes kernel  >= 6.2. Therefore, we are forced to use the older binary `kernel.panic_on_warn=1` enabling which does not allow setting numerical limits.

Could potentially still enable forcing reboot after a single kernel panic using `kernel.panic=-1` to be in full compliance with the KSPP. Thoughts?

## Changes

Set `sysctl kernel.panic_on_warn=1`.

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,9 +43,10 @@ Kernel space:
 
 - Restrict kernel profiling and the performance events system to `CAP_PERFMON`.
 
-- Force the kernel to panic on "oopses" that can potentially indicate and thwart
-  certain kernel exploitation attempts. Optional - Force immediate reboot on the
-  occurrence of a kernel panic and also set panic limit to one (when using Linux kernel >= 6.2).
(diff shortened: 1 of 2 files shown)


adrelanos:
Could you please undo this in https://github.com/Kicksecure/debug-misc?

PR #279, opened 2024-11-08
Provide network-related hardening options via `sysctl`'s

raja-grewal:
Provide option to drop gratuitous ARP packets.

Observed no issues using the `systcl` in my own testing.

## Changes

Currently commented-out.

Can enable the `systcl` after comprehensive testing.

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #289, opened 2024-12-19
Enable ARP filtering

raja-grewal:
As per https://github.com/Kicksecure/security-misc/pull/279#issuecomment-2496914818.

## Changes

Set `sysctl net.ipv4.conf.*.arp_filter=1`

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,8 +105,7 @@ Networking:
 - Optional - Deny sending and receiving shared media redirects to reduce
   the risk of IP spoofing attacks.
 
-- Optional - Enable ARP filtering to mitigate some ARP spoofing and ARP
-  cache poisoning attacks.
+- Enable ARP filtering to mitigate some ARP spoofing and ARP cache poisoning attacks.
 
 - Optional - Respond to ARP requests only if the target IP address is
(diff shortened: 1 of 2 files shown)


ArrayBolt3:
I may just be getting confused by the legitimately difficult-to-understand documentation for `arp_ignore` in Linux, but I don't see anything this option does that `arp_ignore=2` doesn't already do. Personally though, I'm fine with enabling it though as a form of defense-in-depth (if a bug in Linux causes one option to just not work, the other one may be able to keep the system from becoming vulnerable anyway).

I do worry about how this and `arp_ignore=2` may affect bridged networking under VirtualBox and libvirt. Won't these options prevent the VM from being able to find the IP addresses of any other device on the local network? If so, it may still be worth it to break that, but only if we document how to unbreak it and support doing so.

ArrayBolt3:
I guess the current documentation is clear for a user who know how ARP works in Linux. It's a bit cryptic, but the Wiki can help with that, and like in the other places, I'll want to add some bits to the Wiki for these things anyway.

@adrelanos Looks good to me, should be ready for merge.

PR #291, opened 2024-12-19
Drop gratuitous ARP packets

raja-grewal:
As per https://github.com/Kicksecure/security-misc/pull/279#issuecomment-2496914818.

## Changes

Set `sysctl net.ipv4.conf.*.drop_gratuitous_arp=1`

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,8 +111,8 @@ Networking:
 - Optional - Respond to ARP requests only if the target IP address is
   on-link, preventing some IP spoofing attacks.
 
-- Optional - Drop gratuitous ARP packets to prevent ARP cache poisoning
-  via man-in-the-middle and denial-of-service attacks.
+- Drop gratuitous ARP packets to prevent ARP cache poisoning via
+  man-in-the-middle and denial-of-service attacks.
 
 - Ignore ICMP echo requests to prevent clock fingerprinting and Smurf attacks.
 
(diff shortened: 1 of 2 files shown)


ArrayBolt3:
Doesn't look like it's likely to cause problems in a home use scenario, but what happens in a cloud scenario where you rely on other servers on the local network for certain resources, and those servers sometimes have their MAC addresses change? (I don't know if AWS ever has to change the MAC addresses of things like package update servers and whatnot, but I would assume it could happen if they're shuffling VMs around.) That could cause things to break, possibly quite badly. I can also imagine something like this happening in a corporate scenario where a server is reconfigured and it has to tell all client machines on the network the new MAC address for its IP.

I think for *most* users this is probably fine, since IPs don't just randomly switch MACs very often AFAIK. In an environment where gratuitous ARP is needed, the sysadmin will probably know, and be able to turn this setting off. So I think this is another thing that's good to enable by default, but that we should support disabling, and document how to turn it off.

ArrayBolt3:
The in-file documentation is fine to me, but I'd also like to document it in the wiki so people who are like "what on earth, why can't my workstation find my server anymore" can do a web search or wiki search and (hopefully) find it. I can add that to my list of things to do 

@adrelanos I consider this ready to merge.

PR #305, opened 2025-05-15
Prevent erroneous "Login blocked after [negative number] attempts" errors

DMHalford:
For root, faillock appears to always* return an empty string (i.e. no table headers are present), yielding a zero-initialized pam_faillock_output_count and thus resulting in the calculation of a negative failed_login_counter value.

This can cause erroneous errors of the form "ERROR: Login blocked after [negative number] attempts" during sudo-ing and screen unlocking.

This commit modifies the initialization of failed_login_counter such that it cannot be negative and prevents the display of these incorrect warnings.

* Only rudimentary local tests were conducted

This pull request changes...

## Changes
failed_login_counter initialization in /usr/libexec/security-misc

## Mandatory Checklist

## Optional Checklist

- [ ] I have reviewed and updated any documentation if relevant

Fixes #
Erroneous pam-info warnings during sudo-ing and screen unlocking.

### DIFF
--- a/usr/libexec/security-misc/pam-info
+++ b/usr/libexec/security-misc/pam-info
@@ -163,9 +163,9 @@ pam_faillock_output_count="$(echo "$pam_faillock_output" | wc -l)"
 ## example pam_faillock_output_count:
 ## 4
 
-## Do not count the first two informational textual output lines
-## (starting with "user:" and "When").
-failed_login_counter=$(( pam_faillock_output_count - 2 ))
+## Do not count the first two informational textual output lines (starting with "user:" and "When") if present,
+## but ensure failed_login_counter is not set to a negative value.
+failed_login_counter=$( [ $(( pam_faillock_output_count - 2 )) -gt 0 ] && echo $(( pam_faillock_output_count - 2 )) || echo "0" )
 
 ## example failed_login_counter:
 ## 2
--- a/usr/libexec/security-misc/pam-info
+++ b/usr/libexec/security-misc/pam-info
@@ -164,7 +164,7 @@ pam_faillock_output_count="$(echo "$pam_faillock_output" | wc -l)"
 ## 4
 
 ## Do not count the first two informational textual output lines (starting with "user:" and "When") if present,
-## but ensure failed_login_counter is not set to a negative value.
+## whilst ensuring failed_login_counter is not set to a negative value.
 failed_login_counter=$( [ $(( pam_faillock_output_count - 2 )) -gt 0 ] && echo $(( pam_faillock_output_count - 2 )) || echo "0" )
 
 ## example failed_login_counter:


adrelanos:
Will need some refactoring that I'll apply on top.

adrelanos:
Done.

I couldn't really reproduce running `faillock` while not showing its header.

```
user:
When                Type  Source                                           Valid
```

How did you manage to do that? If we could disable the header, then the code could be simplified.

DMHalford:
Thanks for the merge, @adrelanos.

Truth be told, I'm not entirely sure why it was happening, or why it wasn't caught earlier in the script during:

`if [ "$pam_faillock_output" = "" ]; then
   true "$0: no failed login"
   exit 0
fi`

... shortly after the faillock invocation. I'll create a Kicksecure VM this evening and see if I can determine the steps to reproduce.

DMHalford:
Hi again, @adrelanos.

I've done a little debugging and have figured out what was happening. You were correct - faillock *always* outputs the two initial header lines (with no apparent means of disabling it), providing that a valid tally directory is supplied on the command line. (My "rudimentary local tests" overlooked that prerequisite - apologies for my lack of diligence here.)

The PR I submitted actually works-around symptoms of an incompatibility between Kicksecure's included coreutils package and older hardware. The erroneous pam-info errors I was seeing apparently resulted from wc crashing during pam_faillock_output_count initialization:

`pam_faillock_output_count="$(echo "$pam_faillock_output" | wc -l)"`

Obviously, my machine failing to set pam_faillock_output_count to the expected minimum value of two was what resulted in the calculation of a negative failed_login_counter value; it had nothing to do with the output of faillock as I initially presumed.

Running wc under gdb showed that the crash was due to the attempted execution of AVX instructions in the binary, which my (Skylake) CPU (apparently) doesn't support. I've confirmed this by rebuilding coreutils from source and dropping the resulting wc into /usr/bin - your _original_ pam-info script then executed without these 'negative failed_login_counter' bugs presenting.

I'm not sure what you'll want to do with this information, to be honest - perhaps supporting systems _without_ the AVX(2) instruction set(s) would be more effort than it'd be worth, given the rarity of such hardware these days?

I suppose the negativity check for failed_login_counter does make sense, semantically, although it does complicate the code-base for a rather hack-ish work-around for an incompatibility that I dare say would crop up elsewhere in the future for people running Kicksecure on older systems. My OSDev feng shui leaves a lot to be desired, so I'm not really sure what best practices would be here, other than to include both AVX and non-AVX wc builds in the Kicksecure image, coupled with a boot script (containing "grep avx /proc/cpuinfo" perhaps?) to determine the CPU's AVX compatibility and relocate the 'correct' binary into /usr/bin?

DMHalford:
Here's a screenshot of the error message I was seeing, just so you're aware:

pam-info error

DMHalford:
> Thank you!

No worries; it's the least I can do considering that you maintain the whole distro for free!

> But these issues are already (or hopefully getting) fixed in newer versions. (Debian 13 / trixie and above)

Excellent! Glad it's being looked into upstream; one less thing to manage from a Kicksecure perspective 👍
That said, I appreciate the inclusion of a test for wc functionality in the distro's Dev branch(es); probably a smart move. Crashes in a security-centric context could have been disasterous, especially if it flew under the radar like it did somewhat here.

Thanks for your continued efforts with Kicksecure; I might not be the best judge because I'm still dipping my toes into the ocean that is Linux, but I think it's a great distro!

PR #307, opened 2025-05-20
fix(permission-hardener): ssh-agent gets 755 perms

maybebyte:
Replace the commented-out matchwhitelist entry for ssh-agent with an explicit permission entry (755) for /usr/bin/ssh-agent.

When ssh-agent's matchwhitelist entry was commented out in commit 7a5f8b87af, permission-hardener began resetting it to restrictive defaults (744), preventing non-root users from executing ssh-agent. This broke split SSH functionality in Qubes OS for me because I was using Kicksecure in the vault qube, and ssh-agent runs under a non-root user in that configuration (see https://forum.qubes-os.org/t/split-ssh/19060).

As noted in the comment, Debian installs with 2755 permissions as a way to mitigate ptrace attacks, but this rationale doesn't apply due to kernel.yama.ptrace_scope=2 being set in Kicksecure.

Here is a temporary workaround for users in case there is a delay in merging:

```
$ echo '/usr/bin/ssh-agent root root 755' | sudo tee -a /etc/permission-hardener.d/20_user.conf
$ sudo apt-get --reinstall install openssh-client
```

## Mandatory Checklist


ArrayBolt3:
I definitely agree we should do something to fix this, I don't think we realized that ssh-agent was SGID when we did this. I'm not totally sure setting the permissions to just 755 is a good idea though, since even though we *do* have ptrace mostly walled off, a user could disable that setting (and would have a good reason to if they're a developer who needs to use gdb). A user turning off an added security measure in Kicksecure shouldn't result in a system that's less secure than vanilla Debian would be.

According to `sudo find / -group _ssh`, the only thing on the filesystem owned by the `_ssh` group is `/usr/bin/ssh-agent`, and that's set to `2755` by default, so `_ssh` isn't able to write to that file. Thus I think we can just whitelist it and not change its file permissions to strip the SGID bit (or change the owning group to `root`). In this instance, those things aren't hazards, they're actually hardening measures.

Thanks for pointing this out!

maybebyte:
Sure, that line of reasoning makes sense to me. If I'm understanding you correctly, something like `/usr/bin/ssh-agent matchwhitelist` would be the right way to handle this in theory?

However, when I edit `/etc/permission-hardener.d/20_user.conf` or `/usr/lib/permission-hardener.d/25_default_whitelist_ssh.conf` and add that in, I can't actually get this to work for me in local testing. `permission-hardener` still set `/usr/bin/ssh-agent` to 744 permissions. The only way I've been able to get it to work is to add `/usr/bin/ssh-agent exactwhitelist` along with `/usr/bin/ssh-agent 2755 root _ssh` in the next line.

I can see that there are other paths with `matchwhitelist` that work perfectly okay like `pam-tmpdir-helper`. So I think I'm missing something here. Any pointers?

maybebyte:
Hmm. Yeah, I run it the same way as you. Maybe I do have a corrupted permission-hardener state then.

If it works for you, I'll push that change, though I'm not sure how to fix the corruption offhand.

maybebyte:
That fixed it, thank you. :)

ArrayBolt3 (approved):
Looks great to me, works in my testing.

PR #313, opened 2025-08-06
Upgrade `sysctl` settings and docs on kernel panics

raja-grewal:
This draft pull request replaces now redundant `sysctl` settings with their newer, equivalent, and more general counterparts when on Linux kernel >= 6.2.

This continues full compliance with KSPP recommendations.

It also splits the forced immediate reboot on a single kernel panic `sysctl` setting to its own section for clarity. However, the setting remains disabled by default despite this being non-compliant with KSPP recommendations.

Note:
Should ONLY be considered after upgrading to Debian 13 (trixie). Changes have been submitted early to facilitate quicker feedback and review which would ideally enable more swift merging.

## Changes

Redundant `sysctl` settings are removed: `kernel.panic_on_oops=1` and `kernel.panic_on_warn=1`

Upgraded `sysctl` settings are enabled: `kernel.oops_limit=1` and `kernel.warn_limit=1`

These (draft) changes have also been appropriately updated in `debug-misc`.

## Mandatory Checklist

## Optional Checklist


usr/libexec/security-misc/panic-on-oops
```diff
@@ -12,12 +12,14 @@ if [ -f /usr/libexec/helper-scripts/pre.bsh ]; then
...
 fi
 
-## Makes the kernel panic on oopses and warnings. This prevents the
-## kernel from continuing to run a flawed processes. Many kernel
-## exploits will also cause an oops, these settings will make the
-## kernel kill the offending processes.
+## Makes the kernel immediately panic on both oopses and warnings.
+## This prevents the kernel from continuing to run a potentially
+## flawed processes. Many kernel exploits will also cause an oops,
+## these settings will make the kernel kill the offending processes.
```
ArrayBolt3, line 18: > these settings will make the kernel kill the offending processes.

Well... kinda... I guess if you consider a full system crash to be "killing the offending processes", then this is true, but this seems misleading to me. I expect "kill the offending process" to mean that *only* the offending process is killed. "These settings will make the system crash rather than continuing to run after an inconsistent state is hit", or something similar to that, would be better.

usr/libexec/security-misc/panic-on-oops
```diff
@@ -12,12 +12,14 @@ if [ -f /usr/libexec/helper-scripts/pre.bsh ]; then
...
+sysctl kernel.oops_limit=1
+sysctl kernel.warn_limit=1
+
+## Makes the system immediately reboot on the occurrence of a single
+## kernel panic. This is an extreme safety option which also creates
+## a large opening for targeted denial of service attacks.
 #sysctl kernel.panic=-1
```
ArrayBolt3, line 25: Is this really that extreme? Once the kernel panics, the system is effectively dead until it's rebooted anyway, so the only things we gain from not rebooting at this point are:

* If the user happens to be looking at a console where the panic details were dumped, they can see debugging data so they can find out why the kernel panicked, possibly. Arguably that's bad since debugging data can compromise security.
* If the user was trying to shut down when the panic hit, we won't end up immediately powering back on, the user will see the system hang and can hold the power button to shut down fully. Arguably that's bad too, the user can hold the power button to fully shut down even if the system is trying to boot back up, and if the user issued a shutdown command and then ran away, now the system is hung with potentially valuable data in memory, leaving it susceptible to a cold boot attack. A reboot wouldn't eliminate that possibility, but it would at least make things a bit trickier by overwriting some memory with the newly booted system's contents.
* If the system hangs in a graphical session due to a panic, the user will have time to copy any valuable unsaved data that's still on the screen onto something else without losing it. This is valuable, I have many times had applications hang and ended up copying down what I could still see to keep from losing all my unsaved work. I'm not sure it's a compelling reason to do this though.

As for opening the user up to targeted denial of service attacks, this sysctl actually does exactly the opposite - it makes the user *less* susceptible to DoS attacks, because as it stands, panicking a system will cause it to hang *forever* until someone either gets physical access to the system or remotes into it with a hardware KVM or IPMI and resets it that way. I can't find the exact post now, but I believe someone on Reddit reported that a backup server one time got stuck in a panicked state (or similar) for multiple *years* without anyone noticing, meaning no backups were happening for that entire time. A reboot was all it took to return the server to operation. So if we reboot on panic, we're actually making it so that systems that can be made to panic will be somewhat resilient to it. Being able to reboot a victim's system at will isn't great, but being able to hang a victim's system permanently until the victim notices and can get physical access is much, much worse.

I'd suggest we enable this by default when panic-on-oops is enabled, at least on servers (we intend on splitting security-misc up a bit so it can work for servers and desktops differently). On desktops, the ability to recover data left on the screen is valuable enough we might leave this disabled by default, but this is a security vs. usability tradeoff, and I'm not sure the usability benefits of hanging rather than rebooting are worth it here.

usr/lib/sysctl.d/990-security-misc.conf
```diff
@@ -180,20 +180,25 @@ kernel.perf_event_paranoid=3
...
-## The limits are applicable when using Linux kernel >= 6.2 (retained here for future-proofing and completeness).
-##
 #kernel.panic=-1
-#kernel.panic_on_oops=1
-#kernel.panic_on_warn=1
 #kernel.oops_limit=1
 #kernel.warn_limit=1
 
+## Force immediate system reboots on the occurrence of a single kernel panic.
+## This is an extreme safety option which also creates a large opening for targeted denial of service attacks.
```
ArrayBolt3, line 193: If we choose to enable reboot-on-panic by default, the wording here will need to change too.

ArrayBolt3 (changes requested):
I like it, I have a couple of suggestions for a settings change and a wording change that might make sense to apply during the Trixie port though.

usr/libexec/security-misc/panic-on-oops
ArrayBolt3: I'd say lets do it now. Greg Kroah-Hartman's argument is *very* good and I agree with it, but it's specifically in relation to panic_on_warn, regardless of what behavior is taken after a panic. On servers, panic_on_warn might be a bad idea for some for the reasons Greg laid out, but if we're going to use it anyway, it's better to reboot than to hang.

ArrayBolt3 (approved):

ArrayBolt3:
> By forcing instant reboots what we have done is at least begin wiping some memory contents that were previously frozen in place.

This makes me wonder if some form of RAM wiping that happens on bootup might be useful...

PR #237, opened 2024-07-15
Disable some Intel PMT kernel modules

raja-grewal:
Disable some Intel Platform Monitoring Technology Telemetry (PMT) kernel modules.

Disabling was first suggested in Issue https://github.com/Kicksecure/security-misc/issues/224.

## Changes

Add some Intel PMT modules to the list of disabled kernel modules.

Create `disabled-intelpmt-by-security-misc`.

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #236, opened 2024-07-15
Disable more Intel ME kernel modules

raja-grewal:
Disable more Intel Management Engine (ME) kernel modules.

## Changes

Add some Intel ME modules to the list of disabled kernel modules.

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #234, opened 2024-07-15
Disable more kernel modules

raja-grewal:
Reduces attack surface by expanding the list of disabled kernel modules relating to file systems, GPS, network file systems, network protocols/drivers, Thunderbolt, and some miscellaneous drivers. Also provides option to disable more Bluetooth modules.

Applies some suggestions in Issue https://github.com/Kicksecure/security-misc/issues/224.

## Changes

Updated `security-misc.maintscript`.

Moved some previously blacklisted modules to the disabled list.

Replaces`disabled-vivid-by-security-misc` with a more general `disabled-miscellaneous-by-security-misc` that can be used for other modules.

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #230, opened 2024-07-11
Refactor `/etc/modprobe.d/*`

raja-grewal:
Refactor kernel module blacklisting and disabling for clarity, ease-of-use, and future-proofing.

In preparation for future planned changes.

If this PR is approved, I plan on immediately introducing many new additions to our existing module blacklisting and disabling. The splitting of the configuration is necessary to improve readability and not result in a single bewilderingly lengthy configuration file.

## Changes
Splits `/etc/modprobe.d/30-security-misc.conf` into two distinct configuration files, one for blacklisting and one for disabling. Improved documentation of relevant files. 

There are no changes to the actual functionality of the code.

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #340, opened 2025-11-21
Refresh `/etc/modprobe.d/*` blacklisting and disabling

raja-grewal:
This pull request updates our list of blacklisted and disabled kernel modules.

The last major updates were done by us around July 2024 in https://github.com/Kicksecure/security-misc/pull/230, https://github.com/Kicksecure/security-misc/pull/232, https://github.com/Kicksecure/security-misc/pull/234, https://github.com/Kicksecure/security-misc/pull/236, https://github.com/Kicksecure/security-misc/pull/237, https://github.com/Kicksecure/security-misc/pull/238, and https://github.com/Kicksecure/security-misc/pull/245. There were also numerous updates by others not using the GitHub PR system and so harder to link.

This current refresh updates documentation, relocates a few things under better headings, moves some modules from blacklisted to disabled, and adds more disabled modules.

The biggest change is incorporating more disabled file systems form secureblue's [config](https://github.com/secureblue/secureblue/blob/live/files/system/usr/lib/modprobe.d/secureblue.conf). These have all been used for a lengthy period sometime and so should be considered free of any obvious breakages.

## Changes

Primarily disables more uncommon and legacy file systems.

## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/etc/modprobe.d/30_security-misc_disable.conf#security-misc-shared
+++ b/etc/modprobe.d/30_security-misc_disable.conf#security-misc-shared
@@ -17,7 +17,7 @@
 ## 1. Hardware:
 
 ## Bluetooth:
-## Can disable Bluetooth to reduce the attack surface due to its long history of security vulnerabilities.
+## Disable Bluetooth to reduce the attack surface due to its long history of security vulnerabilities.
 ## Replaced with a privacy and security preserving default Bluetooth configuration for better usability.
 ##
 ## https://en.wikipedia.org/wiki/Bluetooth#History_of_security_concerns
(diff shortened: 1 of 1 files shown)


ArrayBolt3:
Some things that could use discussion:

* I don't think we should consider the disabling of `intel_rapl_msr` part of the MSR disablement. `intel_rapl_msr` is AFAICT just a driver for RAPL that happens to use MSRs under the hood instead of TPMI. It doesn't actually expose those MSRs to the end user. We might still want to disable this module though, as RAPL's exposure of detailed power information can be used in side-channel attacks (one of which was known by Intel for four *years* before publication, see https://www.intel.com/content/www/us/en/developer/articles/technical/software-security-guidance/advisory-guidance/running-average-power-limit-energy-reporting.html).
* `isst_if_mbox_msr` also looks like something that just uses MSRs under the hood but probably doesn't expose MSR write functionality directly. I'm not sure if we want to disable this one either.

Some things that need fixed before or after merge:

* Most of the blacklisted filesystems make sense to me, but `9p` cannot be blacklisted, as it's needed for KVM shared folders to work in Whonix (at least with the current documentation for how to set up shared folders). See https://www.whonix.org/wiki/KVM#Shared_Folder
* The new script `disabled-cpumsr-by-security-misc` needs to be explicitly listed in `debian/security-misc-shared.install` due to the new split package architecture.

ArrayBolt3 (commented):
Merged into my arraybolt3/trixie branch, thank you!

PR #341, opened 2025-12-11
Provide the option to enable AMD SEV-ES and SEV-SNP

raja-grewal:
This pull request provides the options to enable two extensions of AMD Secure Encrypted Virtualization (SEV):
- SEV-ES (Encrypted State) extends SEV by encrypting each guests virtual CPU register state during VM exits, and
- SEV-SNP (Secure Nested Paging) extends SEV by activating hardware-level memory integrity.

As per suggested in https://github.com/Kicksecure/security-misc/pull/338#issuecomment-3588000749 by @ArrayBolt3.

## Changes

There are no changes to the functionality of the codebase.

Provide the disabled by default options:
```
kvm_amd.sev_es=1
kvm_amd.sev_snp=1
```

## Mandatory Checklist

- [x] Legal agreements accepted. By contributing to this organisation, you acknowledge you have read, understood, and agree to be bound by these these agreements:

[Terms of Service](https://www.kicksecure.com/wiki/Terms_of_Service), [Privacy Policy](https://www.kicksecure.com/wiki/Privacy_Policy), [Cookie Policy](https://www.kicksecure.com/wiki/Cookie_Policy), [E-Sign Consent](https://www.kicksecure.com/wiki/E-Sign_Consent), [DMCA](https://www.kicksecure.com/wiki/DMCA), [Imprint](https://www.kicksecure.com/wiki/Imprint)

## Optional Checklist
The following items are optional but might be requested in certain cases.

- [x] I have tested it locally
- [x] I have reviewed and updated any documentation if relevant
- [ ] I am providing new code and test(s) for it

PR #343, opened 2025-12-24
Enable SME and SEV in supported AMD CPUs

raja-grewal:
This pull request enables SME and SEV in supported AMD processors.

As per discussed in https://github.com/Kicksecure/security-misc/pull/338 and https://github.com/Kicksecure/security-misc/pull/341.

This has also been suggested by me and approved in https://github.com/secureblue/secureblue/pull/1631.

## Changes
Sets the following kernel boot parameters:
```
mem_encrypt=on
kvm_amd.sev=1
vm_amd.sev_es=1
vm_amd.sev_snp=1
```
## Mandatory Checklist

## Optional Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,11 +46,11 @@ configuration file and significant hardening is applied to a myriad of component
 
 - Restrict kernel profiling and the performance events system to `CAP_PERFMON`.
 
-- Force the kernel to immediately panic on both "oopses" (which can potentially indicate
-  and thwart certain kernel exploitation attempts) and kernel warnings in the `WARN()` path.
-
 - Force immediate system reboot on the occurrence of a single kernel panic, reducing the
   risk and impact of denial-of-service attacks and both cold and warm boot attacks.
+
+- Force the kernel to immediately panic on both "oopses" (which can potentially indicate
+  and thwart certain kernel exploitation attempts) and kernel warnings in the `WARN()` path.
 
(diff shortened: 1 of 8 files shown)


ArrayBolt3:
This is a bit scary to me. While it does appear that the only *really* affected devices are Raven Ridge APUs (since they require the IOMMU to be in passthrough mode), there are very likely users still using hardware from 2018 with Kicksecure in the wild, so I don't think we can easily discount the number of users this could affect (especially since so many of them showed up in 2021 at https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=994453 when Debian got AMD SME enabled by default). As that bug report shows, the results of this can range anywhere from "framebuffer driver used instead of amdgpu" to "early boot kernel panic, system becomes impossible to use and one can't even see the panic message". If we enable this, it will affect all kernels, so a user who doesn't know what they're doing can't simply boot into an older kernel to recover. They have to know that AMD SME is the problem and disable it in the GRUB command line, which most users probably won't know and will have a hard time figuring out on their own.

If there was a way for us to detect at runtime whether a Raven Ridge GPU was in use or not, and enable this option if not, I would be a lot more comfortable with this. That would require some facility to detect extra hardening measures that the running system supports though, which we don't really have yet. @adrelanos Maybe that would be something we could add?

PR #357, opened 2026-04-01
Fix fm-shim-backend D-Bus name registration and quoting

assisted-by-ai:
## Summary
This PR improves the robustness and security of the fm-shim-backend D-Bus service by fixing shell argument quoting issues in the build script and preventing the service from running in a degraded state when another process owns the D-Bus name.

## Key Changes

- **Build script quoting**: Fixed potential word-splitting issues by storing `pkg-config` output in arrays and properly expanding them with `"${array[@]}"` syntax instead of unquoted command substitution
- **D-Bus name registration**: Added `DBUS_NAME_FLAG_DO_NOT_QUEUE` flag to prevent the service from being queued when the D-Bus name is unavailable
- **Error handling**: Changed behavior when another process owns the D-Bus name from a warning with degraded operation to a fatal error, preventing the service from running in an insecure state
- **Systemd service**: Added `NotifyAccess=main` to the service unit to properly handle systemd readiness notifications

## Implementation Details

The changes ensure that:
1. The build process correctly handles pkg-config flags that may contain spaces or special characters
2. The fm-shim-backend service fails fast if it cannot acquire the required D-Bus name, rather than running with reduced functionality
3. The systemd service properly validates readiness notifications from the main process only

https://claude.ai/code/session_01HJPNe7hjMrPvPtuJmPBhNa

### DIFF
--- a/agents/fm-shim-security.md
+++ b/agents/fm-shim-security.md
@@ -1,0 +1,36 @@
+# fm-shim Security Notes
+
+## Architecture
+
+- **Backend** (`fm-shim-backend.c`): C daemon that claims `org.freedesktop.FileManager1`
+  D-Bus name, receives ShowFolders/ShowItems/ShowItemProperties calls, forks a
+  Python frontend with the URI arguments and systemd's user environment.
+- **Frontend** (`fm_shim_frontend.py`): PyQt5 dialog that validates URIs, resolves
+  paths, shows a confirmation prompt, then opens directories via `gio launch`.
+- **Build script** (`build-fm-shim-backend`): Compiles the C backend with
+  extensive GCC hardening flags.
+- **Service** (`fm-shim.service`): systemd user service, Type=notify.
+
+## Rejected Review Suggestions (Kicksecure/security-misc#357)
+
+### pkg-config quoting in build script
+
+pkg-config output is designed to be word-split by shells. Quoting it via bash
+arrays just reimplements word splitting manually. The unquoted
+`$(pkg-config ...)` is intentional. Upstream has a NOTE comment explaining this.
+Do not re-propose this change.
+
+### DBUS_NAME_FLAG_DO_NOT_QUEUE
+
+The D-Bus name queueing behavior is intentional. The IN_QUEUE case withholds
+`READY=1` (triggering a systemcheck warning) while remaining in the queue to
+reclaim the name ASAP. Exiting immediately would leave the name permanently
+unprotected — a worse outcome. Do not re-propose this change.
+
+## Accepted Threat Model
+
+- Local DoS is out of scope.
+- systemd user manager environment is trusted.
+- Directory replacement / symlink swaps are out of scope (frontend documents this).
+- Success replies are returned even on frontend failure (spec safety).
+- Python >= 3.13.5 required.
--- a/agents/security.md
+++ b/agents/security.md
@@ -1,0 +1,12 @@
+# Security Review Notes
+
+## General Guidelines
+
+- Check `./agents/` for component-specific notes before reviewing or commenting.
+- Upstream repository: Kicksecure/security-misc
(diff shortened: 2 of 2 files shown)


usr/lib/systemd/user/fm-shim.service#security-misc-shared
```diff
@@ -8,6 +8,7 @@ StartLimitBurst=3
 
 [Service]
 Type=notify
+NotifyAccess=main
```
ArrayBolt3: Accepted.

usr/libexec/security-misc/build-fm-shim-backend#security-misc-shared
```diff
@@ -44,14 +44,19 @@ gcc_hardening_options+=(
...
-  $(pkg-config --cflags dbus-1) \
-  $(pkg-config --cflags libsystemd) \
+  "${pkg_cflags_dbus[@]}" \
+  "${pkg_cflags_systemd[@]}" \
   /usr/src/security-misc/fm-shim-backend.c \
   -o /usr/bin/fm-shim-backend \
-  $(pkg-config --libs dbus-1) \
-  $(pkg-config --libs libsystemd) \
+  "${pkg_libs_dbus[@]}" \
+  "${pkg_libs_systemd[@]}" \
```
ArrayBolt3: I don't see the point of this. This basically is just implementing word splitting ourselves, rather than using Bash's built-in word splitting. If we have to do word splitting either way, built-in seems better.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -332,7 +332,7 @@ int main(void) {
 
   dbus_name_request_rslt = dbus_bus_request_name(dbus_conn,
     "org.freedesktop.FileManager1",
-    DBUS_NAME_FLAG_REPLACE_EXISTING,
+    DBUS_NAME_FLAG_REPLACE_EXISTING | DBUS_NAME_FLAG_DO_NOT_QUEUE,
```
ArrayBolt3: But... we want to queue, so that even if we can't grab the name immediately, we take it over as soon as we can.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -345,14 +345,8 @@ int main(void) {
...
-  case DBUS_REQUEST_NAME_REPLY_IN_QUEUE:
-    warnx("Requested 'org.freedesktop.FileManager1' name from D-Bus, but was placed in queue.");
-    warnx("This may be a security risk! Please report this bug!");
-    /*
-     * sd_notify(0, "READY=1") is intentionally omitted here, so that
-     * systemcheck will show a warning about the system not being ready
-     */
-    sd_notify(0, "STATUS=Running, security issue detected");
+  case DBUS_REQUEST_NAME_REPLY_EXISTS:
+    errx(1, "Another process already owns 'org.freedesktop.FileManager1' name and did not allow replacement!");
```
ArrayBolt3: I think Claude misunderstood the security risk here. The security risk is that fm-shim-backend wasn't able to immediately grab the name, not that fm-shim-backend continues running even if it can't immediately grab the name. If fm-shim-backend exits immediately, it will *never* be able to grab the name, which is a worse security risk than placing itself in the queue and waiting until it can grab the name. The existing code will result in a systemcheck warning (because the service never tells systemd that it's fully initialized), which is good, and it *also* plugs potential security holes as soon as it possibly can, which is also good. This should stay as-is.

ArrayBolt3 (commented):
Mostly rejected, rationale below.

adrelanos:
Superseded by https://github.com/assisted-by-ai/security-misc/pull/7

PR #358, opened 2026-04-01
Improve fm-shim-backend security and D-Bus error handling

assisted-by-ai:
## Summary
This PR enhances the security and robustness of the fm-shim-backend service by implementing file descriptor cleanup in child processes, improving D-Bus error handling, and adding systemd hardening options.

## Key Changes

**File Descriptor Management:**
- Added automatic cleanup of inherited file descriptors (except stdin/stdout/stderr) in child processes by scanning `/proc/self/fd` before executing the frontend process
- Prevents accidental leakage of sensitive file descriptors to child processes

**D-Bus Improvements:**
- Changed D-Bus name request to use `DBUS_NAME_FLAG_DO_NOT_QUEUE` flag, preventing the service from queuing if the name is already taken
- Converted queued name request from a warning to a fatal error, treating it as a security issue
- Added proper D-Bus error replies for invalid object paths, interfaces, and methods instead of silently ignoring them
- Added cleanup of D-Bus connection and error data structures on exit

**Systemd Hardening:**
- Added security-focused systemd service options:
  - `NoNewPrivileges=yes` - Prevents privilege escalation
  - `ProtectSystem=strict` - Read-only access to system directories
  - `ProtectHome=read-only` - Read-only access to home directory
  - `PrivateTmp=yes` - Private temporary directory
  - `RestrictNamespaces=yes` - Restrict namespace creation
  - `RestrictRealtime=yes` - Disable real-time scheduling
  - `MemoryDenyWriteExecute=yes` - Prevent executable memory allocation
  - `LockPersonality=yes` - Lock process personality

## Implementation Details
- Uses `opendir()` and `readdir()` to enumerate file descriptors safely
- Validates file descriptor numbers using `strtol()` with proper error checking
- Maintains compatibility by preserving the directory file descriptor during enumeration
- D-Bus error replies are only sent when the client expects a reply

https://claude.ai/code/session_01CgrrFhhVWtqtfhvJdEcBWm

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,31 @@
+# Agent Guidelines for security-misc
+
+## fm-shim
+
+### D-Bus name acquisition
+
+The fm-shim-backend intentionally uses `DBUS_NAME_FLAG_REPLACE_EXISTING`
+**without** `DBUS_NAME_FLAG_DO_NOT_QUEUE`. When the name cannot be acquired,
+the process warns and omits `READY=1` so that systemcheck alerts the user.
+Hard-failing (via `errx`) removes the service entirely, which is worse than
+running in a degraded-but-monitored state. Do not change this to a fatal error.
+
+### Systemd sandboxing on services that spawn user-facing apps
+
+Do not add restrictive systemd sandboxing directives (`ProtectHome`,
+`MemoryDenyWriteExecute`, `ProtectSystem=strict`, etc.) to the fm-shim
+service. These restrictions are inherited by child processes — the file
+manager and anything it launches. `MemoryDenyWriteExecute` breaks JIT,
+`ProtectHome=read-only` breaks normal file manager operations, etc.
+
+### Closing inherited file descriptors in forked children
+
+Use the `close_range(3, ~0U, CLOSE_RANGE_UNSHARE)` syscall (requires
+`#define _GNU_SOURCE` and `#include <linux/close_range.h>`). Do not iterate
+`/proc/self/fd` manually.
+
+### D-Bus error replies
+
+When sending D-Bus error replies in multiple places, extract a shared helper
+function (e.g. `send_error_message_maybe()`) rather than duplicating the
+pattern inline.


usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -332,7 +349,7 @@ int main(void) {
 
   dbus_name_request_rslt = dbus_bus_request_name(dbus_conn,
     "org.freedesktop.FileManager1",
-    DBUS_NAME_FLAG_REPLACE_EXISTING,
+    DBUS_NAME_FLAG_REPLACE_EXISTING | DBUS_NAME_FLAG_DO_NOT_QUEUE,
```
ArrayBolt3: Again, no, this is not what we want to do.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -346,13 +363,7 @@ int main(void) {
...
     break;
   case DBUS_REQUEST_NAME_REPLY_IN_QUEUE:
-    warnx("Requested 'org.freedesktop.FileManager1' name from D-Bus, but was placed in queue.");
-    warnx("This may be a security risk! Please report this bug!");
-    /*
-     * sd_notify(0, "READY=1") is intentionally omitted here, so that
-     * systemcheck will show a warning about the system not being ready
-     */
-    sd_notify(0, "STATUS=Running, security issue detected");
+    errx(1, "Requested 'org.freedesktop.FileManager1' name from D-Bus, but was placed in queue. This is a security risk! Please report this bug!");
```
ArrayBolt3: No, this is not what we want to do. This makes the security risk worse.

usr/lib/systemd/user/fm-shim.service#security-misc-shared
```diff
@@ -10,6 +10,14 @@ StartLimitBurst=3
...
 ExecStart=/usr/bin/fm-shim-backend
 Restart=always
+NoNewPrivileges=yes
+ProtectSystem=strict
+ProtectHome=read-only
+PrivateTmp=yes
+RestrictNamespaces=yes
+RestrictRealtime=yes
+MemoryDenyWriteExecute=yes
+LockPersonality=yes
```
ArrayBolt3: I believe all of these restrictions will be inherited by the child process (the file manager) and any processes it launches, which probably will lead to strange behavior. Even `MemoryDenyWriteExecute=yes` might be problematic if the user launches a terminal from within the file manager, then launches a JIT-compiled applications from there.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -393,6 +404,14 @@ int main(void) {
...
       warnx("Received a D-Bus method call for object path '%s', rather than '/org/freedesktop/FileManager1'!",
         dbus_msg_obj_path);
+      if (dbus_message_get_no_reply(dbus_msg) == FALSE) {
+        DBusMessage *err_reply = dbus_message_new_error(dbus_msg,
+          DBUS_ERROR_UNKNOWN_OBJECT, "Unknown object path");
+        if (err_reply != NULL) {
+          dbus_connection_send(dbus_conn, err_reply, NULL);
+          dbus_message_unref(err_reply);
+        }
+      }
```
ArrayBolt3: Accepted with some refactoring to avoid duplication.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -404,6 +423,14 @@ int main(void) {
...
       warnx("Received a D-Bus method call for interface '%s', rather than 'org.freedesktop.FileManager1'!",
         dbus_msg_iface);
+      if (dbus_message_get_no_reply(dbus_msg) == FALSE) {
+        DBusMessage *err_reply = dbus_message_new_error(dbus_msg,
+          DBUS_ERROR_UNKNOWN_INTERFACE, "Unknown interface");
+        if (err_reply != NULL) {
+          dbus_connection_send(dbus_conn, err_reply, NULL);
+          dbus_message_unref(err_reply);
+        }
+      }
```
ArrayBolt3: Accepted with refactoring.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -422,6 +449,14 @@ int main(void) {
...
       warnx("Received a D-Bus method call for unrecognized method '%s'!",
         dbus_msg_method);
+      if (dbus_message_get_no_reply(dbus_msg) == FALSE) {
+        DBusMessage *err_reply = dbus_message_new_error(dbus_msg,
+          DBUS_ERROR_UNKNOWN_METHOD, "Unknown method");
+        if (err_reply != NULL) {
+          dbus_connection_send(dbus_conn, err_reply, NULL);
+          dbus_message_unref(err_reply);
+        }
+      }
```
ArrayBolt3: Accepted with refactoring.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -433,4 +468,7 @@ cleanup:
     dbus_msg_iface = NULL;
     dbus_msg_method = NULL;
   }
+
+  dbus_connection_unref(dbus_conn);
+  dbus_error_free(&error_data);
```
ArrayBolt3, line 469: Accepted.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -192,7 +193,23 @@ void launch_frontend_process(const char *mode_opt, char **uri_list,
...
+        long fd_num = strtol(fd_entry->d_name, &endptr, 10);
+        if (endptr == fd_entry->d_name || *endptr != '\0') {
+          continue;
+        }
+        if (fd_num >= 3 && fd_num != (long)dir_fd) {
+          close((int)fd_num);
+        }
+      }
+      closedir(fd_dir);
+    }
```
ArrayBolt3: I guess this is safe since even if libdbus uses multiple threads in the background (which hopefully it doesn't but you never know!), `fork()` only gives us a single thread. libdbus should set CLOEXEC on its socket file descriptor from what I can tell, and we don't open anything else, so this shouldn't be strictly necessary, but I guess it's a useful hardening measure, so I'll accept the idea. Implementation-wise though, this should be using the `close_range()` system call, not manual iteration through things in `/proc`.

usr/src/security-misc/fm-shim-backend.c#security-misc-shared
```diff
@@ -32,6 +32,7 @@
 #include <unistd.h>
 #include <assert.h>
 #include <sys/wait.h>
+#include <dirent.h>
```
ArrayBolt3: Unnecessary since I'm using `close_range()` instead.

ArrayBolt3 (commented):
Integrated the useful parts of this into https://github.com/ArrayBolt3/security-misc/commit/af4f229046b8ff64eeea599bc72fe47bd35784fc. Notes below.

ArrayBolt3:
Accepted the AGENTS.md file with modifications in https://github.com/ArrayBolt3/security-misc/commit/b6f5cf1787d2ab12832892ab38b79825a1748e9b.

ArrayBolt3:
@adrelanos Could this be closed since it's already been integrated as much as reasonable?

PR #368, opened 2026-07-15
fix(virusforget): correct --clean abort and --user validation

maybebyte:
## Summary

`virusforget --clean` aborts partway through on exactly the input it exists
to remove, and an invalid `--user` is detected but not acted on. These are
two small, independent error-handling fixes to `usr/libexec/security-misc/virusforget`.

## Changes

- **Guard `restore_file` against a missing backup.** Root cause: an
  extraneous file (present in `$HOME`, absent from the committed backup) is
  deleted and then unconditionally passed to `restore_file`, which runs `cp`
  from a backup path that does not exist. Under `set -e` this aborts the
  entire `--clean` run on the first such file, leaving the system partially
  cleaned. The fix returns early when no backup exists — deleting the
  extraneous file is the correct action and is already done by the caller,
  which keeps a forensic copy under `dangerous/`.
- **Exit when `--user` does not exist.** The nonexistent-user branch printed
  a warning to stdout but, unlike every sibling validation branch, neither
  wrote to stderr nor exited, so a mistyped `--user` proceeded against a
  bogus `/home/<user>`. It now writes to stderr and `exit 1`.

## Testing

- `bash -n usr/libexec/security-misc/virusforget` — no syntax errors.
- Manually traced both walk paths (the direct `process_files` call chain and
  the `process_folders` `find | while` subshell). Because the script uses
  `set -e` without `errtrace`, the subshell path aborts via `set -e`
  propagation on the failed pipe rather than the `ERR` trap, but both paths
  stop on the first extraneous file pre-fix and complete after the guard.

## Notes for reviewers

- The two commits are independent and can be reviewed/reverted separately.
- The `restore_file` guard is transparent to its other two callers
  (missing-file and modified-file), which always have a backup present.
- Behavior-only diff (9 insertions, 1 deletion); no reformatting.

### DIFF
--- a/usr/libexec/security-misc/virusforget#security-misc-shared
+++ b/usr/libexec/security-misc/virusforget#security-misc-shared
@@ -297,6 +297,13 @@ unexpected_file() {
 }
 
 restore_file() {
+   ## An extraneous file has no baseline backup to restore from; the caller
+   ## has already removed it, which is the correct action. Return early so the
+   ## 'cp' below does not fail on the missing backup and abort the entire
+   ## '--clean' run (via 'set -e') on the first extraneous file encountered.
+   if [ ! -e "$full_path_backup" ]; then
+      return 0
+   fi
    if [ "$test_mode" = "true" ]; then
       echo "Simulate restoring file... $full_path_original" >&2
       echo cp --no-dereference --archive "$full_path_backup" "$full_path_original"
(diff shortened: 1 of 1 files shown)


maybebyte:
Used Claude Code to find this and assist, but manually reviewed.

ArrayBolt3:
I'm not really happy with the way extraneous files are handled (we shouldn't be calling `restore_file` on those at all), but virusforget looks like it needs refactored / rewritten to work well and match our existing coding style, so I don't think that's a blocker. Merged in https://github.com/ArrayBolt3/security-misc/commit/7d1553c5136942c38c59d4d5c8cec673e2ba1f06.

PR #372, opened 2026-07-15
docs: correct section numbering in config comments

maybebyte:
## Summary

The comment-based section numbering in two hardening config files has drifted out of sync with the sections they actually contain, which misleads anyone navigating the files. This corrects both.

## Changes

- `usr/lib/sysctl.d/990-security-misc.conf`: the header states the file is divided into 5 sections, but the list immediately below enumerates 6 (through `6. CD-ROM/DVD`) — update the count to 6.
- `etc/modprobe.d/30_security-misc_disable.conf`: the Networking section header reads `## 2. Networking`, duplicating the File Systems section; the file's own table of contents and the following `## 4. Miscellaneous` header confirm it should be `## 3. Networking`.

## Testing

- Comment-only change; no automated tests apply. Each corrected number was cross-checked against the section list and table of contents within the same file.


ArrayBolt3:
Merged into my arraybolt3/trixie branch: https://github.com/ArrayBolt3/security-misc/commit/f3a060ed528a66f3524206a22108b59e3de67ef2

PR #373, opened 2026-07-15
fix(remove-system.map): continue past undeletable files

maybebyte:
## Summary

`System.map` files map kernel symbols to addresses, which is handy for exploitation, so this script shreds them. The problem is that it gave up on the first file it couldn't delete, leaving copies in the other locations on disk.
## Changes

- Removed the `[ -w ]` check that gated deletion. Writable does not necessarily mean removable — append-only files return true even though they cannot be deleted. Now we just try `shred` and check whether it worked.
- The old code called `exit 0` in the else branch of that check, which ended the whole run on the first file it rejected.
- Added `chattr -ia` before each attempt, since we determined `System.map` files shouldn't carry immutable or append-only attributes, and that they would possibly complicate kernel package removal.
- Failures now go into a `failed_list` array and get printed after the loop.
- Iterate over the `system_map_list` array instead of word-splitting a string. Cleaner.

## Testing

No automated tests cover this script, so I tested it by hand. I copied both the old and new versions into a temp tree with the four globs rewritten to point inside it, checked the rewrite took before every run, and ran them unprivileged. The old version stops at the first unwritable file and leaves the rest on disk; the new version removes what it can and reports what it couldn't.

### DIFF
--- a/usr/libexec/security-misc/remove-system.map#security-misc-shared
+++ b/usr/libexec/security-misc/remove-system.map#security-misc-shared
@@ -12,31 +12,44 @@ fi
 
 shopt -s nullglob
 
-system_map_location="/boot/System.map* /usr/src/*/System.map* /lib/modules/*/*/System.map* /System.map*"
+system_map_list=( /boot/System.map* /usr/src/*/System.map* /lib/modules/*/*/System.map* /System.map* )
 
-counter=0
-for filename in ${system_map_location} ; do
-   counter=$(( counter + 1 ))
-done
-
-if [ "$counter" -ge "1" ]; then
+if (( ${#system_map_list[@]} > 0 )); then
    echo "INFO: Deleting system.map files..."
 fi
 
+failed_list=()
+
 ## Removes the System.map files as they are only used for debugging or malware.
-for filename in ${system_map_location} ; do
-   if [ -f "${filename}" ]; then
-      if [ -w "${filename}" ]; then
-         ## 'shred' with '--verbose' is too chatty. (7 lines)
-         shred --force --zero -u "${filename}"
-         echo "INFO: removed '${filename}'"
-      else
-         echo "NOTE: Cannot delete '${filename}' - read-only. For details, see: https://www.kicksecure.com/wiki/security-misc#system_map"
-         exit 0
-      fi
+for filename in "${system_map_list[@]}"; do
+   if [[ ! -f "${filename}" ]]; then
+      continue
    fi
+   ## The immutable and append-only file attributes ('chattr +i', 'chattr +a')
+   ## prevent deletion. Clear them. A System.map file has no legitimate
+   ## reason to carry either and they would possibly complicate kernel
+   ## package removal.
+   chattr -ia -- "${filename}" &>/dev/null || true
+
+   ## 'shred' with '--verbose' is too chatty. (7 lines)
+   ## 'shred' stderr is suppressed because failure is expected on read-only
+   ## file systems. Failures are reported below instead.
+   if shred --force --zero -u -- "${filename}" 2>/dev/null; then
+      echo "INFO: removed '${filename}'"
+      continue
+   fi
+   failed_list+=( "${filename}" )
 done
 
-if [ "$counter" -ge "1" ]; then
+if (( ${#failed_list[@]} > 0 )); then
+   for filename in "${failed_list[@]}"; do
+      echo "NOTE: Could not remove '${filename}'."
+   done
+   echo "NOTE: For details, see: https://www.kicksecure.com/wiki/security-misc#system_map"
+elif (( ${#system_map_list[@]} > 0 )); then
    echo "INFO: Done. Success."
 fi
+
+## Deliberately exits 0 even on partial failure. A non-zero exit code would fail
+## the kernel package's postinst hook and the oneshot unit on every boot.
+exit 0


adrelanos:
If the first match cannot be deleted what's the point of attempting to delete others?

maybebyte:
@adrelanos Let me see if I understand what you're saying here. You're saying that, if even one `System.map` file cannot be removed, it is pointless to try to delete the other ones?

Since we are running this service as root, semantics of `[ -w ...]` mean that we are essentially checking if it is on a read-only filesystem or immutable, if I understand correctly.

My thought process was that it isn't necessarily better to abort and keep other copies around. Better to reduce the total number of info leaks, even if it's not perfect. I didn't want to make invasive changes to try and ensure purity on a first pass.

I also find it a little surprising on the user end. If I have a tool that removes `System.map` files, what I would expect as the contract is that it does its best to do that, rather than aborting early. Does that reasoning hold up from your side or am I missing something?

ArrayBolt3:
I think the assumption was that if one file couldn't be deleted, surely the other ones couldn't be either, since the filesystem was read-only. This is because `[ -w file ]` on a read-only file as root states returns 0, i.e. the file is writable despite being "read-only" even to root. There is an exception though; if the file is immutable (`chattr +i`), then it will show as read-only even to root.

I can see the logic behind erasing everything we can rather than aborting, but I wonder if it would be worth using `chattr -i` on each file before attempting to delete. It would override a user who was trying to preserve the file, but if the user wanted to preserve the file that bad, they could just copy it somewhere outside of /boot. Doing a `chattr +i` on one of these files would possibly complicate kernel removal anyway, so it seems to me like these should never be immutable and thus removing the immutable bit if it is present would be the right way to go, in addition to doing this.

tl;dr: I'll probably merge this and add a `chattr -i` to things.

adrelanos:
> I think the assumption was that if one file couldn't be deleted, surely the other ones couldn't be either, since the filesystem was read-only.

Indeed. Because this only known to be happening in some App Qubes, where `/lib/modules` isn't writeable.

`exit 0` -> `continue` makes little sense. If anything, the `exit 0` could be removed without replacement. The effect is the same.

Old output (existing version):

```
INFO: Deleting system.map files...
NOTE: Cannot delete '/lib/modules/.../build/System.map' - read-only. For details, see: https://www.kicksecure.com/wiki/security-misc#system_map
```

New output (when using `exit 0` -> `continue` or removing `exit 0?`) would become_

```
INFO: Deleting system.map files...
NOTE: Cannot delete '/lib/modules/.../build/System.map' - read-only. For details, see: https://www.kicksecure.com/wiki/security-misc#system_map
NOTE: Cannot delete '/lib/modules/.../source/System.map' - read-only. For details, see: https://www.kicksecure.com/wiki/security-misc#system_map
INFO: Done. Success.
```

So the output is worse.

That can be fixed but my thinking was it's more code that in fact doesn't effectively accomplish anything more.

maybebyte:
Hey, thanks for the responses. I'm dealing with an NVMe SSD that is failing at the moment and am writing this off of an old spare computer that doesn't have all my dev tooling set up.

I think `chattr -i` makes sense. ~I believe a file with `chattr +a` would also cause `[ -w ...]` to return failure in these conditions, but that is pretty uncommon and I don't see a clear reason why a `System.map` would be marked append-only.~ EDIT: this is untrue, I tested locally and it does interfere with removal, but it does NOT affect the `[ -w ...]` return code.

Yeah, fair point about the `continue` being unneeded.

I was thinking in terms of the output that we could potentially collect failed files in an array and report them to the user at the end. Then link to the wiki once at the bottom. Or log one line for each failed attempt during iteration instead of storing in an array. Either way, I can squelch the success line since it is misleading. Open to thoughts about that.

I'd be willing to write the code for this once I've dealt with this other less fun stuff IRL, if it's something you all would be interested in. If not, that's cool too. I respect this project and want to help out.

ArrayBolt3:
@maybebyte We're definitely interested in this :) Thanks for your submissions, they're quite helpful.

maybebyte:
@adrelanos @ArrayBolt3

Hey, I finally replaced the drive and set up my dev environment again (managed to restore all of it by imaging the drive with `ddrescue`, carving out the LUKS partition from that image, creating it as a loop device, decrypting, mounting the LUKS container, etc etc... it's a crazy story that involved me finding out 15 minutes before a meeting at $DAYJOB that my computer didn't work, but I'll spare you the details haha).

In any case, I've force-pushed a rewrite and updated the PR description.

Here is what I did:

- Removed the `[ -w ]` check entirely. Writable does not necessarily mean removable. For instance, append-only files return true even though they cannot be deleted. Now we just try `shred` and check whether it worked.
- Added `chattr -ia` to the removal loop before `shred`, since we determined `System.map` files shouldn't have append-only or immutable attributes.
- Iterate over a `system_map_list` Bash array instead of word-splitting a string. Cleaner and not influenced by `IFS`.
- Added a `failed_list` Bash array to collect all failed attempts to remove `System.map` files.
- We now iterate through the `failed_list` Bash array after the removal loop and print all results.
- We only print the wiki link if `failed_list` length > 0.
- We only claim success if `failed_list` length is not > 0 and `system_map_list` length > 0.

Output should now look something like:

```
INFO: Deleting system.map files...
NOTE: Could not remove '/lib/modules/<version>/build/System.map'.
NOTE: Could not remove '/lib/modules/<version>/source/System.map'.
NOTE: For details, see: https://www.kicksecure.com/wiki/security-misc#system_map
```

The two NOTE lines above are really one file reached under two names (`source/` is a symlink to `build/`). I played around with deduplicating by device and inode, but decided against it. It was ~15 lines and a `stat` per failure to save one line of output. I can add it back if desired.

Oh, and the script now ends with an explicit `exit 0`. It's just for documentation purposes, since the `if`/`elif` above it already returns 0. We can drop it if you want.

ArrayBolt3:
Will re-review, sorry for the long delay. Been working on some very intensive refactoring and code review over the last month.

ArrayBolt3:
Merged into my branch in https://github.com/ArrayBolt3/security-misc/commit/97f78d8262aeea854cef3d654ab6aeb57e1f08ec and polished a bit in https://github.com/ArrayBolt3/security-misc/commit/aa458bac395d6fd450c749d0934f4a63d4a76c17. Thank you!

PR #376, opened 2026-07-16
fix(pam): pass profile names to pam-auth-update --remove

maybebyte:
## Summary

The `prerm`/`postrm` scripts try to disable this package's PAM profiles on
removal, but they pass the package name to `pam-auth-update --remove`, which
matches profile *filenames* — so the early-disable was silently a no-op. On
every removal, `/etc/pam.d/common-*` kept referencing helper scripts dpkg had
already deleted; `block-unsafe-logins` is a Priority 1100 `requisite`, so all
authentication failed until postrm's reconciliation ran — permanently if the
postrm failed first (ArrayBolt3's reproduction above).

## Changes

- **Root cause:** `pam-auth-update --remove` matches config **filenames**
  (readdir of `/usr/share/pam-configs/` — see `debian/local/pam-auth-update`
  in the [pam source](https://sources.debian.org/src/pam/latest/debian/local/pam-auth-update/)).
  An unknown target such as `security-misc-shared` is accepted without
  validation and simply never matches.
- **`prerm`** (on `remove` only): pass the 8 actual profile filenames,
  unguarded. This is the man-page-documented purpose of `--remove` ("before
  the modules they reference are removed from disk") and the ecosystem norm:
  libpam-mount, sssd, and fprintd all use this exact shape (unguarded,
  `$1 = remove` only), and debputy auto-generates it. If pam-auth-update were
  somehow unrunnable, the removal aborts — failing closed instead of
  recreating the dangling-requisite lockout.
- **`postrm`**: replace the inert `--remove <pkg>` with a plain
  `pam-auth-update --package` reconciliation in **all** postrm states,
  guarded by `command -v` (Policy §7.2: no dependency guarantee at postrm
  time). With profile files still on disk (`upgrade`/`abort-*`)
  reconciliation is a no-op, so nothing is stripped mid-transaction — this
  addresses the review question about `failed-upgrade`/`abort-install`/
  `abort-upgrade`.

## Testing

Everything below ran in `debian:stable-slim` against a .deb built from this
revision (the real profiles, the referenced helper scripts, these exact
maintainer scripts) and can be re-run from the descriptions alone.

- Root cause is directly observable: `pam-auth-update --package --remove
  security-misc-shared` changes nothing, while `--remove` with the actual
  profile filenames removes the entries. With this PR, `prerm remove` strips
  all 8 profiles from `common-*` (including an `--enable`d console-lockdown)
  while the helper scripts and profile files are still on disk, and
  `dpkg -i` → `dpkg -r` → `dpkg -P` exit 0 with no reference to
  `/usr/libexec/security-misc` left anywhere in `/etc/pam.d` at any point.
- Failure injection with ArrayBolt3's pre.bsh trick (fail only the postrm):
  with the current code, removal fails and `common-auth` is left pointing at
  the deleted `block-unsafe-logins` helper — the reported lockout state; with
  this PR, removal still fails but `/etc/pam.d` is already clean. Forcing
  prerm to fail *after* its `--remove` instead makes dpkg run
  `postinst abort-remove`, the package stays installed, and the outcome is
  exactly the behavior change disclosed under Notes below.
- Every postrm invocation dpkg can make (`upgrade`, `failed-upgrade`, both
  `abort-install` argument forms, `abort-upgrade`, `disappear`, `purge`)
  exits 0 and leaves `common-*` untouched while the profile files are
  present. `postrm purge` also exits 0 with pam-auth-update removed from
  PATH, and `prerm remove` without pam-auth-update fails closed as intended.
  `bash -n` passes on both scripts; shellcheck reports nothing on the
  changed lines.

## Notes for reviewers

- **Behavior change vs the old inert code** (also disclosed in the commit
  message): an effective `--remove` drops the 8 names from
  `/var/lib/pam/seen`. If a removal is aborted after prerm ran (dpkg then
  calls `postinst abort-remove`, whose reconciliation runs), the seven
  `Default: yes` profiles are re-enabled — including any an administrator had
  deliberately disabled — and `console-lockdown-security-misc`
  (`Default: no`) stays off until re-enabled; reinstalling does not restore
  it, since preinst's `--enable` is behind a `do_once` stamp. On all
  *completed* paths (remove, purge, upgrade, remove→reinstall) the end state
  is identical to before this change. The same side-effect class exists in
  sssd and libcap2.
- **Caveat:** the 8-name list in `prerm` must stay in sync with
  `debian/security-misc-shared.install` and `usr/share/pam-configs/`; a
  forgotten name is still swept by the postrm reconciliation within the same
  dpkg run.
- **Pre-existing limitation (unchanged):** with local modifications inside
  the pam-auth-update managed block of `common-*`, pam-auth-update declines
  to act and exits 0, so neither the old nor the new calls change anything on
  such systems.

### DIFF
--- a/debian/security-misc-shared.postrm
+++ b/debian/security-misc-shared.postrm
@@ -16,7 +16,11 @@ true "
 "
 
 ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/11
-pam-auth-update --package --remove "$DPKG_MAINTSCRIPT_PACKAGE"
+## Debian Policy 7.2: there is no guarantee that dependencies (here
+## libpam-runtime) are available when postrm runs; skip gracefully then.
+if command -v pam-auth-update >/dev/null 2>&1; then
+   pam-auth-update --package
+fi
 
 rm -f /etc/sysctl.d/30_security-misc_aslr-mmap.conf
 
--- a/debian/security-misc-shared.prerm
+++ b/debian/security-misc-shared.prerm
@@ -16,7 +16,20 @@ true "
 "
 
 if [ "$1" = remove ]; then
-   pam-auth-update --package --remove "$DPKG_MAINTSCRIPT_PACKAGE"
+   ## 'pam-auth-update --remove' matches pam-config FILENAMES under
+   ## /usr/share/pam-configs/, not the package name. Disable while the helper
+   ## scripts under /usr/libexec/security-misc/ still exist. Keep in sync with
+   ## security-misc-shared.install and usr/share/pam-configs/.
+   ## See pam-auth-update(8).
+   pam-auth-update --package --remove \
+      block-unsafe-logins-security-misc \
+      console-lockdown-security-misc \
+      faillock-preauth-security-misc \
+      mkhomedir-security-misc \
+      pam-abort-on-locked-password-security-misc \
+      umask-security-misc \
+      unix-faillock-security-misc \
+      wheel-security-misc
 fi
 
 true "INFO: debhelper beginning here."


adrelanos:
Have you been able to reproduce this bug - what bug exactly? And was this PR tested and known to fix the issue?

ArrayBolt3:
Looking at the pam-update-auth documentation, I think this change is *probably* good, though I need to make sure that the arguments being checked for in prerm make sense.

@adrelanos The bug is:

* If security-misc-shared's prerm succeeds,
* but then the postrm fails before `pam-auth-update --package` is called,
* the configuration files under `/etc/pam.d` will still contain PAM configuration from security-misc, but the scripts that configuration depends on will have been removed from disk. PAM will therefore break entirely.
* There may be other edge cases where a PAM module is left around when the files it depends on are removed by an upgrade, which would come with similar consequences.

The solution is to remove PAM configuration from `/etc/pam.d` before security-misc's files are removed from disk. This is documented in `pam-auth-update`'s manual page:

> `pam-auth-update --remove` should be used to remove profiles from the configuration before the modules they reference are removed from disk, to ensure that PAM is in a consistent and usable state at all times during package upgrades or removals.

An easy way to reproduce the issue is to make `/usr/libexec/helper-scripts/pre.bsh` improperly terminate the postrm script, and only the postrm script, by adding the following code to the file immediately after the shebang:

```bash
if [ "${DPKG_MAINTSCRIPT_PACKAGE}" = 'security-misc-shared' ] \
   && [ "${DPKG_MAINTSCRIPT_NAME}" = 'postrm' ] \
   && [[ "${1}" =~ ^(remove|purge) ]]; then
   exit 1
fi
```

Then run `sudo apt remove security-misc-shared`. The removal will fail (`installed security-misc-shared package post-removal script subprocess returned error exit status 1`), and if you attempt to run `sudo -i` thereafter, it will break:

```
[sysmaint ~]% sudo -i
/usr/libexec/security-misc/block-unsafe-logins failed: exit code 2
sudo: PAM authentication error: Unknown error -1
sudo: a password is required
zsh: exit 1     sudo -i
```

So if this PR doesn't break anything and does fix the above, it's good.

debian/security-misc-shared.postrm
```diff
@@ -16,7 +16,19 @@ true "
...
 ## https://forums.whonix.org/t/is-security-misc-suitable-for-hardening-bridges-and-relays/8299/11
-pam-auth-update --package --remove "$DPKG_MAINTSCRIPT_PACKAGE"
+## Reconciliation safety net; the authoritative early-disable is in prerm. By the
+## time postrm runs dpkg has deleted this package's /usr/share/pam-configs/ files,
+## so a plain reconciliation drops the now-file-less profiles from common-*.
+## Plain '--package' only (no '--remove <profiles>'): on upgrade/abort states the
+## config files are still present and '--remove' would strip live hardening
+## mid-transaction, so those states are deliberately not matched here.
+case "$1" in
+   remove|purge|disappear)
```
ArrayBolt3: Should we match `failed-upgrade`, `abort-install`, and `abort-upgrade` too? (See https://www.debian.org/doc/debian-policy/ch-maintainerscripts.html#summary-of-ways-maintainer-scripts-are-called, specifically the subsection about how the `postrm` script may be called.) I don't think this will be problematic for any of these (or for `upgrade`, for that matter), and it might not be redundant in any of those situations. (It is redundant for `upgrade` though, since the postinst will have already called `pam-auth-update`.)

ArrayBolt3 (changes requested):

debian/security-misc-shared.postrm
maybebyte: I ended up reviewing this and decided to drop the case statement so it runs in every postrm state. A plain `pam-auth-update --package` only drops profiles whose files are missing from `/usr/share/pam-configs`. In upgrade and the failed/abort states our files are still on disk, so it changes nothing. And in an abort-install where they never landed, dropping the entries is the right cleanup anyway. In remove/purge/disappear they're gone, so it cleans up. The command -v guard is for states where libpam-runtime itself may already be gone.

Tested every postrm invocation dpkg can make against a built package in debian:stable-slim: all exit 0, and none touch common-* while the files are present.

maybebyte:
Yes, reproduced. The bug is this: `pam-auth-update --remove` takes profile filenames (whatever is in `/usr/share/pam-configs/`), not package names. In terms of proof, there is the `readdir/%removals` handling in `debian/local/pam-auth-update` in the pam source, or trying it interactively works too: `pam-auth-update --package --remove security-misc-shared` changes nothing, `pam-auth-update --package --remove umask-security-misc` removes the profile (a plain `pam-auth-update --package` re-enables it afterwards, since it's `Default: yes` and freshly unseen). So the `--remove` in our prerm and postrm has been a silent no-op all along, and on removal dpkg deletes the helper scripts while `/etc/pam.d/common-*` still references them. `block-unsafe-logins` is a requisite line, so authentication breaks until postrm's reconciliation cleans up — permanently if the postrm dies first, which is what ArrayBolt3's transcript shows.

Tested using ArrayBolt3's `pre.bsh` reproduction from this thread, against a built package in `debian:stable-slim`. Current code: removal fails and `common-auth` is left pointing at the deleted `/usr/libexec/security-misc/block-unsafe-logins`. On this PR's code with that `pre.bsh` change, removal still fails, but `/etc/pam.d` is already clean, because prerm disabled the profiles while the files were still on disk. Without that `pre.bsh` change, install → remove → purge all exit 0 and leave no reference to `/usr/libexec/security-misc` anywhere in `/etc/pam.d`.

ArrayBolt3 (approved):
This has been merged into my arraybolt3/trixie branch in https://github.com/ArrayBolt3/security-misc/commit/a85c5b57045e95a840c4059e66d8296b8e51c155, and the comments tweaked a bit in https://github.com/ArrayBolt3/security-misc/commit/827e5ee1f8a250ba8b04e08c91892b1887e83d61 to remove some LLM verbosity. Thank you!

